PR #2, opened 2014-11-09
Added flags that allow ipv6 databases to be used

whyscream:
I've added the flags that make the plugin support ipv6 city and country databases.

I was unable to get the rspec test suite running (even in pristine mode), probably due to the separate repository setup, but I've verified that this works using a simple test config and the getting started setup. Test databases were retrieved from maxmind using the `geoip-database-contrib` package in ubuntu trusty (multiverse).

### DIFF
--- a/lib/logstash/filters/geoip.rb
+++ b/lib/logstash/filters/geoip.rb
@@ -79,9 +79,9 @@ class LogStash::Filters::GeoIP < LogStash::Filters::Base
     geoip_initialize = ::GeoIP.new(@database)
 
     @geoip_type = case geoip_initialize.database_type
-    when GeoIP::GEOIP_CITY_EDITION_REV0, GeoIP::GEOIP_CITY_EDITION_REV1
+    when GeoIP::GEOIP_CITY_EDITION_REV0, GeoIP::GEOIP_CITY_EDITION_REV1, GeoIP::GEOIP_CITY_EDITION_REV1_V6
       :city
-    when GeoIP::GEOIP_COUNTRY_EDITION
+    when GeoIP::GEOIP_COUNTRY_EDITION, GeoIP::GEOIP_COUNTRY_EDITION_V6
       :country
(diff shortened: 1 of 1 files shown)


jsvd:
I tried adding simple ipv6 test using this patch and a google ipv6 dns and couldn't make it work:

```ruby
  describe "ipv6 query" do
    config <<-CONFIG
      filter {
        geoip {
          source => "ip"
        }
      }
    CONFIG

    sample("ip" => "2001:4860:4860::8888") do
      insist { subject }.include?("geoip")
    end
  end
```

not sure if the ip isn't in the database or something else isn't working..

SamSaffron:
getting errors galore with this ... debugging

SamSaffron:
actually it seems to be working @jsvd  you got to load up the ipv6 db for it to work

SamSaffron:
+1 please please merge this

jsvd:
@SamSaffron interesting, then maybe this PR should include the fetching of the ipv6 dat files since it won't work without them. I'll try to get that working quickly

whyscream:
At least in the stable ubuntu package, the geocitylite db file is included in the logstash package (in the vendor/ folder). SO not part of this plugin, or fetched dynamically.

jsvd:
@whyscream yep, I meant adding that to the list of stuff to package when creating the plugin gem, like in #8
Actually, @whyscream if you don't mind I'll close this follow up this feature on #8 since it already includes the tests for ipv6, it's only necessary to update the links in the json.

whyscream:
@jsvd Whatever makes you happy, as long as we get ipv6 support released ;)

SamSaffron:
@jsvd #8 does not have a CLA, probably not worth touching unless you can organize that ... FYI this is how I work around the issue now 

https://gist.github.com/SamSaffron/953d3b0de2662b8af205

suyograo:
Will be fixed via https://github.com/logstash-plugins/logstash-filter-geoip/pull/64

PR #9, opened 2015-01-23
Removed unnecessary escape characters

fhopf:
There are some backslashes in there that are rendered to the html as well. Adding the backslash to the filter config will cause the field to not be created.

I haven't tested the resulting Ruby code but I can see that in other files the underscores are not escaped.


jsvd:
there are other backslashes in the :target field documentation, could you change those too please?

PR #10, opened 2015-01-28
Add vendor.json

purbon:
So the dependencies are properly installed during installation time at logstash

### DIFF
--- a/vendor.json
+++ b/vendor.json
@@ -1,0 +1,10 @@
+[
+    {
+        "url": "http://logstash.objects.dreamhost.com/maxmind/GeoLiteCity-2013-01-18.dat.gz",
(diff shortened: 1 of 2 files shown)


vendor.json
```diff
@@ -0,0 +1,5 @@
+[ {"url": "http://logstash.objects.dreamhost.com/maxmind/GeoLiteCity-2013-01-18.dat.gz", 
+  "sha1": "15aab9a90ff90c4784b2c48331014d242b86bf82" },
```
jsvd: INCOHERENT INDENTATION OF THE JSON DOCUMENT

purbon: I will fix that, but this should be considered for now just a PoC as there are still missing issues with the Bundler implementation to get the vendored files.

Issue #19, opened 2015-04-03
"fields" doesn't work as expected for "location"


volter:
What is happening:

If you don't specify "longitude" and "latitude" in "fields", "location" is not created, even if you requested it in "fields".

Why is it happening:

"location" is constructed from "longitude" and "latitude" property of the event.

Expected behaviour:

Construct "location" regardless whether "longitude" and/or "latitude" are specified in "fields".

jordansissel:
Agreed on the behavior being a bug.

jordansissel:
Fixed by #20/#22, thank you! :)

PR #20, opened 2015-04-07
Generate "location", no matter if lon and lat are desired to be in the r...

volter:
...esult or not

- Solves #19 

Could be more efficient, but it does the job!

### DIFF
--- a/lib/logstash/filters/geoip.rb
+++ b/lib/logstash/filters/geoip.rb
@@ -119,6 +119,10 @@ class LogStash::Filters::GeoIP < LogStash::Filters::Base
     geo_data_hash = geo_data.to_hash
     geo_data_hash.delete(:request)
     event[@target] = {} if event[@target].nil?
+    if geo_data_hash.key?(:latitude) && geo_data_hash.key?(:longitude)
+      # If we have latitude and longitude values, add the location field as GeoJSON array
+      geo_data_hash[:location] = [ geo_data_hash[:longitude].to_f, geo_data_hash[:latitude].to_f ]
+    end
     geo_data_hash.each do |key, value|
       next if value.nil? || (value.is_a?(String) && value.empty?)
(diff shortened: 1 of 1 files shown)


untergeek:
I like the solution.  Can you please add some spec tests to [spec/filters/geoip_spec.rb](https://github.com/logstash-plugins/logstash-filter-geoip/blob/master/spec/filters/geoip_spec.rb) to validate the new behavior?

jordansissel:
I wrote some tests for this. My tests fail without this patch and pass with it.

The failures before patching:
```
  1) LogStash::Filters::GeoIP location field when latitude field is excluded behaves like location field should have a location field
     Failure/Error: expect(event["[geoip][location]"]).not_to(be_nil)
       expected: not nil
            got: nil
     Shared Example Group: "location field" called from ./spec/filters/geoip_spec.rb:139
     # ./spec/filters/geoip_spec.rb:133:in `(root)'

  2) LogStash::Filters::GeoIP location field when longitude field is excluded behaves like location field should have a location field
     Failure/Error: expect(event["[geoip][location]"]).not_to(be_nil)
       expected: not nil
            got: nil
     Shared Example Group: "location field" called from ./spec/filters/geoip_spec.rb:144
     # ./spec/filters/geoip_spec.rb:133:in `(root)'

  3) LogStash::Filters::GeoIP location field when both latitude and longitude field are excluded behaves like location field should have a location field
     Failure/Error: expect(event["[geoip][location]"]).not_to(be_nil)
       expected: not nil
            got: nil
     Shared Example Group: "location field" called from ./spec/filters/geoip_spec.rb:149
     # ./spec/filters/geoip_spec.rb:133:in `(root)'
```

jordansissel:
My pr (#22) builds on this PR and adds tests. Going to close this in favor of #22. Thank you for fixing this, @volter! :)

Issue #21, opened 2015-05-13
IPv6 ASN lookup throws exception: #<NoMethodError: undefined method `to_hash' for ...


jordansissel:
(Migrated from https://github.com/elastic/logstash/issues/3235; filed by @jannae)

--- 
I'm at a bit of a loss, so any help is appreciated. I'm getting a few errors that seem to have brought my ELK instance to it's knees. It stops dead on these, each time I attempt to restart or update the config in any way. Not sure what to make of it, as all my research has led me to previously corrected issues. 

It seems related to this:
 
<https://logstash.jira.com/browse/LOGSTASH-1394>

Though this is indicated as having been corrected. :(

Also, if you need to review my configs they are here: https://github.com/jannae/logstash-configs

This may or may not be related to my recent upgrade to 1.5.0 rc4, it's tough to tell, except that this is a brand new problem for me, so, seems plausible.

```
{:timestamp=>"2015-05-12T22:11:44.302000-0400",
    :message=>"Exception in filterworker",
        "exception"=>#<NoMethodError: undefined method `to_hash' for "*AS18144 Energia Communications,Inc.":String>,
        "backtrace"=>["/opt/logstash/vendor/bundle/jruby/1.9/gems/logstash-filter-geoip-0.1.9/lib/logstash/filters/geoip.rb:119:in `filter'",
        "/opt/logstash/vendor/bundle/jruby/1.9/gems/logstash-core-1.5.0.rc4-java/lib/logstash/filters/base.rb:162:in `multi_filter'",
        "org/jruby/RubyArray.java:1613:in `each'",
        "/opt/logstash/vendor/bundle/jruby/1.9/gems/logstash-core-1.5.0.rc4-java/lib/logstash/filters/base.rb:159:in `multi_filter'",
        "(eval):10100:in `cond_func_383'",
        "org/jruby/RubyArray.java:1613:in `each'",
        "(eval):10094:in `cond_func_383'",
        "(eval):3085:in `filter_func'",
        "/opt/logstash/vendor/bundle/jruby/1.9/gems/logstash-core-1.5.0.rc4-java/lib/logstash/pipeline.rb:219:in `filterworker'",
        "/opt/logstash/vendor/bundle/jruby/1.9/gems/logstash-core-1.5.0.rc4-java/lib/logstash/pipeline.rb:156:in `start_filters'"],
    :level=>:error}
{:timestamp=>"2015-05-12T22:11:44.349000-0400",
[...]

Issue #12, opened 2015-01-28
Natively support IPv6 address lookups


tylerjl:
The geoip filter works great for IPv4 lookups, but there's a couple of IPv6 addresses on the internet, and we'd like to be able to look those up, too. :+1:

[Maxmind provides IPv6 databases](http://dev.maxmind.com/geoip/legacy/geolite/), and the filter provides a `database` option, so while this can be done manually by dropping the IPv6 database onto a machine and adding logic to use the correct database, it would certainly be nice to at least bundle it along with v4 database in the plugin so that one could change the database setting to the IPv6 one explicitly after doing some regex catching:

    if [ip] =~ /:/ {
        geoip {
            database => "GeoLiteCityv6.dat"
            ...
        }
    } else {
        geoip {
            ...
        }
    }

*or*, the filter could handle it transparently if it's feeling ambitious and detect the address type within the plugin itself and make life even better for everyone.

untergeek:
Migrating https://logstash.jira.com/browse/LOGSTASH-1100 from JIRA.

archimedesdev:
Hi, just curious if there's been any progress or traction on this request. Thanks!

untergeek:
It's important to us, but please pardon the delays while we prioritize the release of Logstash 1.5.

TinLe:
I believe this is already fixed.   Did the PR not get merged?

https://github.com/logstash-plugins/logstash-filter-geoip/pull/8

untergeek:
Not yet, no.

Deshke:
any update on this ?

goacid:
Ping

SamSaffron:
@untergeek any news here, its a 1 line change ... :(

untergeek:
We're busy refactoring all plugins to do work differently for 2.0.  Also, a refactor of the plugin to use the newer database format is pending.  This change will probably not be added until then.

hanej:
+1

IngaFeick:
+1

sdsunjay:
+1

killafunkinmofo:
+1

panda87:
+1

pemontto:
:+1:

userguy:
any update on This .. The Geoip is not able to look at Maxmind database for ipv6

SamSaffron:
We use a patched up version to get this working, something along the lines of:

```
 ruby {
      code => "
[...]

PR #8, opened 2014-11-21
fix ipv6 support for geoip

42wim:
Simple fix to support the IPv6 databases in the filter.

Fixes 
- https://logstash.jira.com/browse/LOGSTASH-1100

and partly
- https://logstash.jira.com/browse/LOGSTASH-1124
- https://logstash.jira.com/browse/LOGSTASH-2163

reworked from https://github.com/elasticsearch/logstash/pull/1350

PR #24, opened 2015-05-20
fix ASN IPv6 address lookup crash

colinsurprenant:
solves #21

### DIFF
--- a/spec/filters/geoip_spec.rb
+++ b/spec/filters/geoip_spec.rb
@@ -1,12 +1,35 @@
 require "logstash/devutils/rspec/spec_helper"
 require "logstash/filters/geoip"
 
+ASNDB = ::Dir.glob(::File.expand_path("../../vendor/", ::File.dirname(__FILE__))+"/GeoIPASNum*.dat").first
+
 describe LogStash::Filters::GeoIP do
+
+  describe "ASN db" do
+    config <<-CONFIG
+      filter {
+        geoip {
+          source => "ip"
+          database => "#{ASNDB}"
+        }
+      }
+    CONFIG
+
+    sample("ip" => "1.1.1.1") do
+      insist { subject["geoip"]["asn"] } == "Google Inc."
+    end
+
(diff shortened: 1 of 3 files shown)


spec/filters/geoip_spec.rb
```diff
@@ -1,12 +1,35 @@
 require "logstash/devutils/rspec/spec_helper"
 require "logstash/filters/geoip"
 
+ASNDB = ::Dir.glob(::File.expand_path("../../vendor/", ::File.dirname(__FILE__))+"/GeoIPASNum*.dat").first
```
ph: Any reason why not using a `let` here? It wont be used outside of the test.

colinsurprenant: when I tried it did not work and I did not feel like looking for the reason since this is an "old" style specs that desperately need to be refactored anyway.

ph: it is a namespace issue? Please create an issue to refactor them.

ph:
LGTM, merge.
The semantic is consistent with the NIL value. :shipit:

ph:
Lets create another issue to support IPv6?

colinsurprenant:
I just realized there is, #12

spec/filters/geoip_spec.rb
colinsurprenant: *I don't know*

ph:
@colinsurprenant there is also a PR to add ipv6 support with test, I dont know the state of it. 
https://github.com/logstash-plugins/logstash-filter-geoip/pull/8 ?

colinsurprenant:
yeah, I meant PR #8, #12 is the issue. Don't know the state, will have to review it. In any cases, since `GeoIP` methods seems to occasionally return a `String` on *bad* lookups, checking for the `to_hash` method support should stay to prevent any other corner cases.

colinsurprenant:
merging and publishing.

ph:
Agree merge!

PR #28, opened 2015-06-05
Fix error variable

jhitze:
The logging messages were passing the wrong field name back to user when something went wrong.

I also added to the documentation so that hopefully no one else runs into the issue of not doing the `rake vendor` task.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,12 @@ bundle install
 bundle install
 ```
 
+- Pull down GeoIP database files
+
+```sh
+bundle exec rake vendor
+```
+
 - Run tests
 
 ```sh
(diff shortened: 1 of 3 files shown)


jsvd:
LGTM

jsvd:
@jhitze it seems you signed the CLA with an email but commited this patch [with another](https://patch-diff.githubusercontent.com/raw/logstash-plugins/logstash-filter-geoip/pull/28.patch) could you either sign it with the email that appears in this PR or redo the commits and force push it with the other email?

jhitze:
>.<

Should be fine now.

suyograo:
Internal CLA check is ok

suyograo:
Merged https://github.com/logstash-plugins/logstash-filter-geoip/commit/56ac1af682964436318e307f3385e90a54659c95

PR #35, opened 2015-08-04
Remove \ escapes since they show up in the docs.

evanstoner:
Backslashes show up before underscores in the "target" documentation. 

See https://www.elastic.co/guide/en/logstash/current/plugins-filters-geoip.html#plugins-filters-geoip-target


evanstoner:
@jordansissel CLA signed.

suyograo:
Fixed in docs.

PR #36, opened 2015-09-03
Return StandardError where applicable

SamSaffron:
```
sam@ubuntu logstash % bin/logstash agent -f ../elastic/config/haproxy.conf
The error reported is: 
  uninitialized constant LogStash::Filters::GeoIP::RuntimeException

After change

sam@ubuntu logstash % bin/logstash agent -f ../elastic/config/haproxy.conf
The error reported is: 
  This GeoIP database is not currently supported
sam@ubuntu logstash % 
```

### DIFF
--- a/lib/logstash/filters/geoip.rb
+++ b/lib/logstash/filters/geoip.rb
@@ -84,7 +84,7 @@ class LogStash::Filters::GeoIP < LogStash::Filters::Base
     when GeoIP::GEOIP_ISP_EDITION, GeoIP::GEOIP_ORG_EDITION
       :isp
     else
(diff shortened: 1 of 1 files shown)


SamSaffron:
@untergeek  any news here, this seems rather minor.

untergeek:
This change may not work properly with plans for Logstash 2.0.  The plan is to use APIs to configure, so a STDERR output is not ideal in that case.  Error reporting and all other considerations will be factored in with the rewrites.

kaisecheng:
This is for a legacy Logstash version which is not supported anymore, hence close this PR

PR #40, opened 2015-09-14
WIP Lru cache

andrewvc:


andrewvc:
This provides a nearly 50% speedup on a test dataset. 

```
~/D/logstash-1.5.4 $ sh -c 'sleep 120 && killall -9 java' & bin/logstash -w2 -f complete.conf | pv -l > /dev/null
 927k 0:02:00 [7.73k/s] [                     <=>                                                  ]
fish: Process 40059, 'bin/logstash' from job 2, 'bin/logstash -w2 -f complete.co…' terminated by signal SIGKILL (Forced quit)
'sh -c 'sleep 120 && killall -9…' has ended
~/D/logstash-1.5.4 $ emacs Gemfile
~/D/logstash-1.5.4 $ sh -c 'sleep 120 && killall -9 java' & bin/logstash -w2 -f complete.conf | pv -l > /dev/null
1.57M 0:02:00 [13.1k/s] [                     <=>                                                  ]
fish: Process 40108, 'bin/logstash' from job 2, 'bin/logstash -w2 -f complete.co…' terminated by signal SIGKILL (Forced quit)
'sh -c 'sleep 120 && killall -9…' has ended
```

lib/logstash/filters/geoip.rb
```diff
@@ -88,6 +109,11 @@ def register
     end
 
     @threadkey = "geoip-#{self.object_id}"
+
+    # This is wrapped in a mutex to make sure the initialization behavior of LOOKUP_CACHES (see def above) doesn't create a dupe
+    LOOKUP_CACHE_INIT_MUTEX.synchronize do
+      @lookup_cache = LOOKUP_CACHES[@geoip_type] ||= LruRedux::ThreadSafeCache.new(1000)
+    end
```
ph: make sense!

ph: Instead of an ivar we could use an accessor method?

andrewvc: Can do!

spec/filters/geoip_spec.rb
```diff
@@ -221,6 +221,23 @@
...
+      plugin.register
+    end
+
+    it "should dup the objects" do
+      event = {}
+      alt_event = {}
+      plugin.apply_geodata(geo_data, event)
+      plugin.apply_geodata(geo_data, alt_event)
 
+      event["geoip"].each {|k,v| expect(v).not_to eql(alt_event[k]) }
```
andrewvc: Oops, this is bad. This should be checking object_ids. These values SHOULD match. Uh oh...

jordansissel: I was gonna ask that, but then I remembered that Ruby has like 6 ways to measure equality and maybe rspec's eql used the one that compares object id or something and then my mind wandered onto more pleasant things like cheese and hockey.

lib/logstash/filters/geoip.rb
```diff
@@ -130,16 +154,36 @@ def filter(event)
...
-            when Encoding::ASCII_8BIT; value.force_encoding(Encoding::ISO_8859_1).encode(Encoding::UTF_8)
-            when Encoding::ISO_8859_1, Encoding::US_ASCII;  value.encode(Encoding::UTF_8)
-            else; value
-          end
+                    # I have found strings coming from GeoIP that are ASCII-8BIT are actually
+                    # ISO-8859-1...
+                    when Encoding::ASCII_8BIT; value.force_encoding(Encoding::ISO_8859_1).encode(Encoding::UTF_8)
+                    when Encoding::ISO_8859_1, Encoding::US_ASCII;  value.encode(Encoding::UTF_8)
+                    else; value.dup
+                  end
```
ph: concerning the `#dup`, I think the getter `get_geo_data_for_ip` should return the duped values?

spec/filters/geoip_spec.rb
ph: cheese.

ph: But this is a good question I think rspec's eql is checking `object_id` for that specific case.

ph: Can use `to_hash` and with `match` to compare the actual values.

lib/logstash/filters/geoip.rb
andrewvc: The rationale here was to be more efficient with duplication. The ASCII encoding step makes a copy, so that would cause extra object creation.

lib/logstash/filters/geoip.rb
```diff
@@ -58,9 +64,24 @@ class LogStash::Filters::GeoIP < LogStash::Filters::Base
...
+  #
+  # It is important to note that this config value is global to the geoip_type. That is to say all instances of the geoip filter
+  # of the same geoip_type share the same cache. The last declared cache size will 'win'. The reason for this is that there would be no benefit
+  # to having multiple caches for different instances at different points in the pipeline, that would just increase the
+  # number of cache misses and waste memory.
+  config :lru_cache_size, :validate => :number, :default => 1000
+
   public
   def register
-    require "geoip"
```
colinsurprenant: Do we want to change that strategy where "3rd party" gems are actually required in the `register` method instead of in the module top? we all agree & understand that this is an anti-pattern and we should avoid requiring elsewhere than at module top (there are totally acceptable exceptions) but in our case, the purpose for this, IIRC was to actually requiring all 3rd party gems unless the plugins were actually specified in a config. Maybe this is not a concern anymore, not sure.

lib/logstash/filters/geoip.rb
colinsurprenant: a spec to verify for this should be added and if it does not pass then we simply add the `dup` no?

colinsurprenant: yes, the encode method returns a new string indeed so definitely no `dup` need there.

lib/logstash/filters/geoip.rb
ph: I believe this is an anti-pattern too and we should move all the require at the top of the file and stop  doing lazy require of library. There is no real gain in lazy mode, since the plugins need to be installed and loaded via `Plugin#lookup` to actually load the required files, this also help us track errors more easily.

lib/logstash/filters/geoip.rb
ph: @andrewvc +1 on the efficient part of it.

untergeek:
These LRU Cache additions are awesome, and very straightforward. :+1:

ph:
I am OK with this PR! Test run!

I have looked at the geoip code, there is an option to load the entire dataset into memory, but I think the LRU approach is more memory friendly and will provide an awesome speed boost!

ph:
LGTM!

lib/logstash/filters/geoip.rb
colinsurprenant: Nonetheless I'd like to make sure we understand why this was previously needed and why it is not now, not just "guessing" it should be ok now and change a construct that was purposefully valid at some point. It is certainly the case that it is not required now but I'd like that to not be guess work.

untergeek:
Indeed.  Compressed, the entire dataset is around 20MB, I think.  Loading that into memory could be a bit much.

colinsurprenant:
Did we measure the uncompressed db memory size? 20MB is peanuts and ~100MB would still be peanuts. it could be configurable? LRU lookups should always be faster than even an in-memory db since there is no parsing to do but a combination or LRU & in-memory could be an extreme combo!

untergeek:
```sh
» ls -lh GeoLiteCity-2013-01-18.dat
-rw-r--r-- 1 buh buh 18M Sep 15 10:49 GeoLiteCity-2013-01-18.dat
```
Not sure what it's compressed with, but there it is.

ph:
created this https://github.com/logstash-plugins/logstash-filter-geoip/issues/41

lib/logstash/filters/geoip.rb
andrewvc: Sorry, I have a habit of fixing little things as I go. I'm fairly certain this was just 'the style at the time' (I used to wear an onion on my belt), so I don't think its dangerous.

lib/logstash/filters/geoip.rb
andrewvc: Yep, there already is a spec covering this.

lib/logstash/filters/geoip.rb
colinsurprenant: I'm fairly certain it was **not** *the style at the time* and it was purposely there to avoid mass requiring all 3rd party gems when all the plugins were installed but not configured or something around these lines.

jordansissel: STORY TIME!

The reason Logstash historically put `require "some third party thing"` in the `register` method was because .... computers.

The libraries we used in Logstash (JRuby) were not always available or installed for MRI, and at the time, I used MRI to generate the docs, and because the docs generator calls `eval` on the code :smiley: those top-of-the-file `require` calls would fail because the gems weren't available. Why MRI for docs building (at the time)? Because, way back when, the only "good" markdown libraries for Ruby only worked in MRI. Uuuuugh computers.

So that's the story. `require` in `register` because of markdown and mri and jruby.

Today? I don't think these requirements are necessarily true anymore, though it would be nice to be able to build the docs _without_ needing the runtime dependencies, we can probably hack around that as needed.

+1 on, long term, moving requires to the top of the file.

colinsurprenant: awwwwwwww thanks for the background on this.

PR #42, opened 2015-09-15
Lazy-load geoip gem

andrewvc:
This backs out a change made in #40 that shouldn't have been part of that PR. This will now require 'geoip' on register.


colinsurprenant:
LGTM

PR #43, opened 2015-09-16
Be more defensive about threadsafety

andrewvc:
Some specs were randomly failing, this fixes it by being more defensive about Thread.current

### DIFF
--- a/lib/logstash/filters/geoip.rb
+++ b/lib/logstash/filters/geoip.rb
@@ -28,6 +28,7 @@ class LogStash::Filters::GeoIP < LogStash::Filters::Base
   LOOKUP_CACHES = {}
 
   attr_accessor :lookup_cache
+  attr_reader :threadkey
 
   config_name "geoip"
 
@@ -176,7 +177,7 @@ class LogStash::Filters::GeoIP < LogStash::Filters::Base
     if (cached = lookup_cache[ip])
       cached
     else
-      geo_data = Thread.current[@threadkey].send(@geoip_type, ip)
+      geo_data = Thread.current[threadkey].send(@geoip_type, ip)
       lookup_cache[ip] = geo_data
       geo_data
     end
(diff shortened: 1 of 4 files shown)


lib/logstash/filters/geoip.rb
```diff
@@ -189,4 +181,15 @@ def get_geo_data_for_ip(ip)
...
   end
+
+  def ensure_database!
+    # Use thread-local access to GeoIP. The Ruby GeoIP module forces a mutex
+    # around access to the database, which can be overcome with :pread.
+    # Unfortunately, :pread requires the io-extra gem, with C extensions that
+    # aren't supported on JRuby. If / when :pread becomes available, we can stop
+    # needing thread-local access.
+    if !Thread.current.key?(@threadkey)
+      Thread.current[@threadkey] = ::GeoIP.new(@database)
```
ph: This should be enough no, since `Thread#current` access should be thread safe?

```ruby
 Thread.current[@threadkey] ||= ::GeoIP.new(@database)
```

ph: @andrewvc Not sure how this would fix the issue since we just moved the code closer  to the lookup. Why Thread.current would be different than the `#filter` method?

andrewvc: This came up in a spec where I didn't want to test filter, just the geoip lookup.

jsvd:
is there some way to change a test to always fail and check it's fixed with this? maybe insert a sleep somewhere?

andrewvc:
@jsvd I don't believe so. I think the issue was that that spec worked because previous speaks 'leaked' the info in Thread.current (which is why setting stuff there is dangerous). I suspect it should have failed every time if run without the other tests. I don't think its time-related.

lib/logstash/filters/geoip.rb
ph: OK, yeah we have this kind of issues poping up in a few places, the more I see this issue the more I think we should bring back `#initialize` in the plugins.

Like @jsvd  mentioned a failling test would be cool

andrewvc: Well, with the new defensive code, there should be no way to make it fail, which is a good thing :)

ph: maybe something like 
```
Thread.current[mythread_key]

Comment (andrewvc) [inline lib/logstash/filters/geoip.rb]: Ah, yes, I'll add that in.

**Comment by andrewvc** [comment]
OK, I added a spec for this behavior

**Comment by ph** [comment]
LGTM

**Comment by ph** [comment]
my bad as a reviewer I should have run the test.

```

  1) LogStash::Filters::GeoIP re-initializing thread current DB should initialize the DB on lookup, regardless of thread state
     Failure/Error: Thread.current.thread_variables.delete plugin.threadkey
     NoMethodError:
       undefined method `thread_variables' for #<Thread:0x5cbf9e9f run>
     # ./spec/filters/geoip_spec.rb:258:in `(root)'

Finished in 0.964 seconds (files took 1.96 seconds to load)
```

PR #44, opened 2015-09-16
This check for Thread.current is more portable across JRuby versions

andrewvc:


ph:
Same failures  you have to change the test too?

Have you tested on Jruby 1.7.2x ?
```
Failures:

  1) LogStash::Filters::GeoIP re-initializing thread current DB should initialize the DB on lookup, regardless of thread state
     Failure/Error: Thread.current.thread_variables.delete plugin.threadkey
     NoMethodError:
       undefined method `thread_variables' for #<Thread:0x6831d8fd run>
     # ./spec/filters/geoip_spec.rb:258:in `(root)'

jordansissel:
Thread.current.key? works for me in JRuby 1.7.19  - what does this change fix?

andrewvc:
@jordansissel Apparently `Thread.current.thread_variables.delete` doesn't work on some rubies. I meant to switch the test to just set the value to nil, which meant I couldn't use `key?`

andrewvc:
Ok, pushed new test code that should fix all this weirdness.

jordansissel:
@andrewvc ahh ok that makes sense.

andrewvc:
@jordansissel it's hard to push the right code when the error isn't happening on your own box. Apologies for the confusion :)

ph:
@andrewvc LGTM, test passes.

PR #45, opened 2015-10-01
Inclusion in main branch: Geomapping for internal IPs

jan-tee:
This version has added support for an additional lookup table for IP addresses to geo information to enable geomapping for internal and edge gateway addresses that would otherwise be inaccurate or unavailable.

The approach is to use a JSON file with geoinformation and IP address range information that will be queried before the GeoIP DB and used in its place if there is a match.


kaisecheng:
Thanks for submitting the PR, however, we have no plan to maintain this feature in the master branch.

PR #46, opened 2015-10-07
Logstash-core >= 2.0.0 < 3.0.0

ph:
This changes allow the plugins to work under theses versions:

- 2.0.0.snapshot*
- 2.X 
- 3.0.0.dev.

The plugins wont work under version:
- 3.0.0

Fixes https://github.com/elastic/logstash/pull/3986

### DIFF
--- a/logstash-filter-geoip.gemspec
+++ b/logstash-filter-geoip.gemspec
@@ -20,7 +20,7 @@ Gem::Specification.new do |s|
   s.metadata = { "logstash_plugin" => "true", "logstash_group" => "filter" }
 
   # Gem dependencies
-  s.add_runtime_dependency "logstash-core", "~> 2.0.0.snapshot"
+  s.add_runtime_dependency "logstash-core", ">= 2.0.0", "< 3.0.0"
 
   s.add_runtime_dependency 'geoip', ['>= 1.3.2']
   s.add_runtime_dependency 'lru_redux', "~> 1.1.0"


ph:
@purbon can you review this? This will be part of a mass update I want to do today.

purbon:
the pattern looks reasonable, I tested with 3.0.0.dev and works, but could not test for 2.0.0.pre ( **I'm not sure it will work**). need to be sure this also works.

ph:
@purbon Which branch you are using? You also have to define the plugins in the gemfile with the correct path and the path need to have the patch applied?

purbon:
I testes with master, updating all to 3.0.0.dev, and with latest
upstream/2.0, last one fail on bin/plugin install because of the dash hack.

/purbon

On Wed, 7 Oct 2015 17:36 Pier-Hugues Pellerin <notifications@github.com>
wrote:

@ <https://github.com/purbon>purbon <https://github.com/purbon> Which
branch you are using? You also have to define the plugins in the gemfile
with the correct path and the path need to have the patch applied?

—
Reply to this email directly or view it on
<https://github.com/logstash-plugins/logstash-filter-geoip/pull/46#issuecomment-146235642>
GitHub
<https://github.com/logstash-plugins/logstash-filter-geoip/pull/46#issuecomment-146235642>.

ph:
There is no beta in https://github.com/elastic/logstash/blob/2.0/lib/logstash/version.rb the upstream/2.0 ?

refering to your previous link https://github.com/elastic/logstash/commit/4f87e53d5098de5fe0a04e8c681517512c321abc#diff-ac8bc3252a541708f5514e8a18317c8b

purbon:
I did no say beta, i said dash in the version name vs having something else
in the gemfile.

/purbon

purbon:
Link is actually where the dash was introduced.

purbon:
@ph I tested your changed and in order to work with pre releases we need to have this pattern.

``` s.add_runtime_dependency "logstash-core", ">= 2.0.0.snapshot", "< 3.0.0" ```

otherwise it only works for non pre releases. This pattern will also work with non pre release versions until 3.0.0.

In my case by using the actual upstream/2.0 branch:
* I cleaned up all installed plugins from the Gemfile.
* Removed the Gemfile.lock file.
* Removed ```vendor/bundle```  directory.
(here I have a clean env)
* run rake bootstrap.
* install the build plugin as in this gem. And get this error:

```
purbon-elastic% bin/plugin install /Users/purbon/work/plugins/filters/logstash-filter-geoip/logstash-filter-geoip-2.0.0.gem
Validating /Users/purbon/work/plugins/filters/logstash-filter-geoip/logstash-filter-geoip-2.0.0.gem
Installing logstash-filter-geoip
Plugin version conflict, aborting
ERROR: Installation Aborted, message: Bundler could not find compatible versions for gem "logstash-core":
  In snapshot (Gemfile.lock):
    logstash-core (= 2.0.0.snapshot2)

  In Gemfile:
    logstash-filter-geoip (= 2.0.0) java depends on
      logstash-core (~> 2.0) java

    logstash-core (= 2.0.0.snapshot2) java
```

if building again the plugin with the new pattern, the installation works as expected.

ph:
good catch @purbon  thanks for verifying this.
I'll change it with your suggestion.

ph:
done in a mass update on october 8 9h40 EST

PR #49, opened 2015-10-27
Update geoip.rb

clintongormley:
Updated link to ES docs to point to 1.7


magnusbaeck:
Any reason for keeping the geo_point documentation link pinned to the ES 1.7 documentation? ES 2.0 changed the name of the page which I guess is the reason for the pinning in the first place, but with both 2.0 and 2.1 out the door it would seem reasonable to have this link point to https://www.elastic.co/guide/en/elasticsearch/reference/current/geo-point.html instead.

PR #52, opened 2015-11-12
Fix for issue 50, manifests only when Cabin is set for IO

### LINKED ISSUE #50: Bad geoip lookup returns invalid hash
Please see https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/298#issuecomment-156066999

guyboertje:
fixes #50

### DIFF
--- a/lib/logstash/filters/geoip.rb
+++ b/lib/logstash/filters/geoip.rb
@@ -124,22 +124,25 @@ class LogStash::Filters::GeoIP < LogStash::Filters::Base
 
   public
   def filter(event)
-    
+
     geo_data = nil
 
     geo_data = get_geo_data(event)
 
-    return if geo_data.nil? || !geo_data.respond_to?(:to_hash)
+    # defense against GeoIP code returning something that can't be made a hash
+    return unless geo_data.respond_to?(:to_hash)
 
-    apply_geodata(geo_data, event)
-
-    filter_matched(event)
+    event[@target] = {} if event[@target].nil?
+    geo_data_hash = geo_data.to_hash
+    # don't do anything more if the lookup result is empty
+    if !geo_data_hash.empty?
+      apply_geodata(geo_data_hash, event)
+      filter_matched(event)
+    end
   end # def filter
 
-  def apply_geodata(geo_data,event)
-    geo_data_hash = geo_data.to_hash
+  def apply_geodata(geo_data_hash, event)
     geo_data_hash.delete(:request)
-    event[@target] = {} if event[@target].nil?
     if geo_data_hash.key?(:latitude) && geo_data_hash.key?(:longitude)
       # If we have latitude and longitude values, add the location field as GeoJSON array
       geo_data_hash[:location] = [ geo_data_hash[:longitude].to_f, geo_data_hash[:latitude].to_f ]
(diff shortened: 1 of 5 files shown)


guyboertje:
@andrewvc - Please review (and enjoy)

jsvd:
wow so
1. `get_geo_data` calls `get_geo_data_for_ip` which can raise 2 exceptions.
2. if `get_geo_data_for_ip` works then it returns an object with hierarchy `[GeoIP::City, #<Class:0x782e409f>, Struct, Enumerable, Object, Kernel, BasicObject]`
3. if not, it caches the exceptions and calls `@logger.error` which returns and object with hierarchy `[Hash, JRuby::Generator::Iterators, Enumerable, Object, Kernel, BasicObject]`
4. the filter method then takes the return of `get_geo_data` and does `geo_data.nil? || !geo_data.respond_to?(:to_hash)` which will always return true because both `GeoIP::City` and Hash respond to `to_hash`
5. in the event of @logger.error injecting something weird to its returning hash (like an IO object) we see this problem.

@guyboertje I'm wondering if we should change the check in filter to ensure the return is either a Struct or more precisely a GeoIP::City (though not sure if this is always true)

spec/filters/geoip_spec.rb
```diff
@@ -184,15 +185,46 @@
...
 
       sample("ip" => "~") do
-        expect{
-          subject
-          }.to_not raise_error
+        expect{ subject }.to_not raise_error
+      end
+    end
+
+    context "when IO is used for logging" do
```
jsvd: looking at this test I'm not sure if we should try to work with whatever `@logger.error` gives us and ensure that either we get a proper Struct from `get_geo_data` or consider the call failed.
This way we don't have to go into details like "what if Cabin logger uses a StringIO?"

guyboertje:
wow is correct.  In point 3 did you mean catches and not caches?

jsvd:
yes, edited.

guyboertje:
@jsvd @purbon  - FYI, updated code and test

lib/logstash/filters/geoip.rb
```diff
@@ -124,22 +124,24 @@ def register
...
-    
+
     geo_data = nil
 
     geo_data = get_geo_data(event)
 
-    return if geo_data.nil? || !geo_data.respond_to?(:to_hash)
-
-    apply_geodata(geo_data, event)
+    return unless geo_data.respond_to?(:to_hash)
```
jsvd: from my understanding this will never be executed since we now control the output of get_geo_data and either pump out an empty hash or a geoip struct?

lib/logstash/filters/geoip.rb
```diff
@@ -124,22 +124,24 @@ def register
...
 
-    filter_matched(event)
+    event[@target] = {} if event[@target].nil?
+    geo_data_hash = geo_data.to_hash
+    # don't do anything more if the lookup result is empty
+    if !geo_data_hash.empty?
```
jsvd: an alternative to line 135 + checking if it's empty here you can just check if geo_data != {}

jsvd:
not necessary but if you find the time would be nice to add an extra test that checks:

* if plugin is `geoip { source => "message"  add_tag => "done" }`
* with an event `LogStash::Event.new("message" => "potatoes")`
* the filtered event won't have tag "done"

lib/logstash/filters/geoip.rb
guyboertje: Its protection against the third party GeoIP code every returning something not hashable.

andrewvc: :+1: for defensive programming, but this could really use a comment explaining it.

lib/logstash/filters/geoip.rb
guyboertje: True but as I decided that `apply_geodata` should expect and operate on a hash, I moved the hash generation to the `filter` method.  So I test the hash for empty.

One could easily argue that the `get_geo_data_for_ip` method should return a hash, meaning that we cache the hash.  Indeed the mutation of the hash in lines 144 - 148 should be done once before it is stored in the cache.  No need to do it each time, especially if there are many events referencing the same IP address.

guyboertje:
@jsvd - some replies.  I will add the other test you requested.

lib/logstash/filters/geoip.rb
guyboertje: Sure OK.

Also, @andrewvc please comment on my other comment https://github.com/logstash-plugins/logstash-filter-geoip/pull/52#discussion_r44779884

jsvd: Ok but then it should have the same effect as a failed lookup: event[@target] becomes {} if nil

lib/logstash/filters/geoip.rb
jsvd: :+1:

andrewvc: @guyboertje I really like the proposal for `get_geo_data_for_ip` to return a hash, but it's not the most important thing for me, so make that determination for yourself :)

guyboertje: @andrewvc - Noted.  I have created a new issue for that refactor.

jsvd:
LGTM

PR #56, opened 2015-11-26
use fieldref to properly assign event field data

colinsurprenant:
this change is necessary to support the future Java Event. 

The previous construct 
```ruby
event[@target][key] = foo
```
is actually: 1- fetch field reference `@target` and return value, 2- create `key` in value and assign `foo`. 
this actually assumes the mutability of `event[@target]` by its reference, which work but is an actual implementation side-effect and not a contract. 

The correct way to assign Event field values is by using proper field references
```
event["[#{@target}][#{key}]"]
```

This relates to elastic/logstash#4264 and elastic/logstash#4191

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 2.0.5
+ - Use proper field references
+
 ## 2.0.4
  - Refactor GeoIP Struct to hash conversion to minimise repeated manipulation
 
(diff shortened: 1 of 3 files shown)


jsvd:
LGTM

colinsurprenant:
thanks, will bump version and merge!

colinsurprenant:
pushed version 2.0.5

PR #57, opened 2015-12-10
added option to disable logging in the geoip filter, useful in production environment

ahzamali:
add configuration property 
	nologging => true      # to completely silence geoip pluging

### DIFF
--- a/lib/logstash/filters/geoip.rb
+++ b/lib/logstash/filters/geoip.rb
@@ -45,6 +45,10 @@ class LogStash::Filters::GeoIP < LogStash::Filters::Base
   # this field is an array, only the first value will be used.
   config :source, :validate => :string, :required => true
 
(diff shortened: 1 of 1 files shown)


ahzamali:
I have signed today 15 min ago at 9:27AM 10 Dec, 2015 IST

jordansissel:
Most of the logging you propose to remove isn't logged by default (info and debug logs are not shown by default with Logstash's default settings).  What is the goal that is not solved by changing your desired log level with a flag in logstash? Can you tell me more about your needs?

ahzamali:
The exception log logs the whole event, in my case the events were wrongly filled with malformed IP addresses and at a high rate the logs was filling up the whole disk. 
At lease we need to have a way to disable the printing of the event itself, just the field should be good.

Issue #58, opened 2015-12-22
Feature request: skip local subnetwork IPs when an array is provided 


xo4n:
Would it be possible to skip private IPs when an array of IPs is provided instead of lookup the first IP of the array. It will be a matter of checking the IPs against the subnetworks as defined https://en.wikipedia.org/wiki/Private_network and then iterate to the next item in the array

IngaFeick:
+1

jordansissel:
Hmm, I'm open to this idea. Skipping lookups on all RFC1918 addresses sounds like an OK default behavior - thoughts?

As for having an array, would it be useful to have multiple geoip lookup results? Or do you just want the first non-rfc1918 address looked up?

IngaFeick:
@jordansissel I am currently working on something like this, will submit a PR somewhen this week. I chose to use the first non-private IP for lookup, but I'm open to suggestions :)

jordansissel:
@IngaFeick +1 on your proposal :)

PR #59, opened 2016-01-04
Split ip field if it contains multiple ips, and lookup the first public ip.

IngaFeick:
This adresses issue 58 https://github.com/logstash-plugins/logstash-filter-geoip/issues/58
When containing more than one ip, only the first non private (RFC1918) ip will be used for lookup. This is helpful if you want to remove proxy ips and use the actual client ip. 
If desired, this can be deactivated in the config to restore the previous behaviour.

### DIFF
--- a/lib/logstash/filters/geoip.rb
+++ b/lib/logstash/filters/geoip.rb
@@ -92,7 +92,7 @@ class LogStash::Filters::GeoIP < LogStash::Filters::Base
 
   # list of ip patterns that private ips start with. TODO benchmark if regex would be faster here.
   config :private_ip_prefixes, :validate => :array, :required => false, :default => ["10.", "192.168." ,"172.16.", "172.17.", "172.18.", "172.19.",
-   "172.20.", "172.21.", "172.22.", "172.23.", "172.24.", "172.25.", "172.26.", "172.27.", "172.28.", "172.29.", "172.30.", "172.31."]
+   "172.20.", "172.21.", "172.22.", "172.23.", "172.24.", "172.25.", "172.26.", "172.27.", "172.28.", "172.29.", "172.30.", "172.31.", "127.0.0"]
 
   public
   def register
@@ -105,9 +105,6 @@ class LogStash::Filters::GeoIP < LogStash::Filters::Base
       end
     end
     @logger.info("Using geoip database", :path => @database)
-
-    @logger.debug("Private ip filter settings: filter_private_ips " + (filter_private_ips ? "true" : "false"))
-   
     # For the purpose of initializing this filter, geoip is initialized here but
     # not set as a global. The geoip module imposes a mutex, so the filter needs
     # to re-initialize this later in the filter() thread, and save that access
@@ -171,6 +168,8 @@ class LogStash::Filters::GeoIP < LogStash::Filters::Base
       else 
         ip = ip.first
       end
+    else
+      @logger.debug("Is string")# TODO remove
     end
     return ip
   end
@@ -184,6 +183,8 @@ class LogStash::Filters::GeoIP < LogStash::Filters::Base
         return ip
       end
     end
+    @logger.debug("no public ip found in string " + ip_array.to_s+ ", returning nil") # TODO remove
+    return nil
   end
 
   def is_private(ip)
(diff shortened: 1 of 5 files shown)


lib/logstash/filters/geoip.rb
```diff
@@ -84,6 +84,16 @@ class LogStash::Filters::GeoIP < LogStash::Filters::Base
   # number of cache misses and waste memory.
   config :lru_cache_size, :validate => :number, :default => 1000
 
+  # If the ip field is an array: check which one is not a private ip. The first non private ip will be used.
+  config :filter_private_ips, :validate => :boolean, :required => false, :default => true
+
+  # String by which the ip field should be tokenized
+  config :ip_split_symbol, :validate => :string, :required => false, :default => ","
+
+  # list of ip patterns that private ips start with.
```
jordansissel: Would it be easier to use CIDR notation? 172.16.0.0/12, etc?

lib/logstash/filters/geoip.rb
```diff
@@ -84,6 +84,16 @@ class LogStash::Filters::GeoIP < LogStash::Filters::Base
   # number of cache misses and waste memory.
   config :lru_cache_size, :validate => :number, :default => 1000
 
+  # If the ip field is an array: check which one is not a private ip. The first non private ip will be used.
+  config :filter_private_ips, :validate => :boolean, :required => false, :default => true
+
+  # String by which the ip field should be tokenized
+  config :ip_split_symbol, :validate => :string, :required => false, :default => ","
```
jordansissel: There's a filter which will already split a field (mutate's split), can we reuse that instead of adding a new setting here?

jordansissel: (I am open to being persuaded that having split support within geoip plugin is useful, but need some convincing)

IngaFeick: The problem with the mutate split function is that it works inplace. If, for whatever reason, the user wants to keep the original value of the ip list, that would be impossible. Also I think it's more elegant to provide the functionality as a whole, to make it work standalone without requiring outside conditions. Makes it easier for the user and less error prone.

lib/logstash/filters/geoip.rb
IngaFeick: We thought it would probably be the fastest approach, performance wise, as it allows us to do simple substring comparisons. CIDR is certainly more elegant when it comes to the notation, but I don't expect the users to modify this value, so I didn't put much weight onto that aspect. An other plus, on the other hand, would be that it makes it much easier to add the private ipv6 subnets also, so if you prefer the CIDR notation, then I'd be happy to implement that.

jordansissel: I'd love to have CIDR notation. If performance becomes a problem, you could do an optimization on common cases (/8, /16, /24) to do substring comparison for those. Could even do that for things like /12 -- My hope was for users to see CIDR notation. In the code itself we can do whatever we like :)

IngaFeick: Ok, I'll give it a try and report back in a couple of days :)

IngaFeick: @jordansissel I updated the PR.

IngaFeick: Any news on this? I basically just copied the adress comparison code from the cidr filter and it works fine for us. Let me know if you would like me to change anything else. Thanks :)

Cactusbone:
This is really useful since it allows to simply use the X-Forwarded-For header without modification. It evens handles (with a warning) the "unknown" ip sent by some squid servers with `forwarded_for=no`, instead of dropping the whole log when first ip in array is "unknown" (current behaviour)

Cactusbone:
I'm getting `Invalid IP network, skipping` for `10/8` and `192.168/16` shouldn't those be `10.0.0.0/8` and `192.168.0.0/16` ?

IngaFeick:
@Cactusbone thanks for the feedback, I will look into it.

IngaFeick:
Hey guys, I would appreciate it if you close this PR, because
- it's been a year and the whole thing is quite outdated
- I have come to the conclusion that it would be better to extract the functionality into a separate field, so as to maintain everything easily. Here's the repo: https://github.com/IngaFeick/logstash-filter-forwarded
Cheers and thanks!

PR #63, opened 2016-03-03
Add geoip preload option

ebuildy:
Add geoip preload to make it working with pipeline multithread.

### DIFF
--- a/lib/logstash/filters/geoip.rb
+++ b/lib/logstash/filters/geoip.rb
@@ -99,7 +99,7 @@ class LogStash::Filters::GeoIP < LogStash::Filters::Base
     # not set as a global. The geoip module imposes a mutex, so the filter needs
     # to re-initialize this later in the filter() thread, and save that access
     # as a thread-local variable.
-    geoip_initialize = ::GeoIP.new(@database)
+    geoip_initialize = ::GeoIP.new(@database, { :preload => true })
 
     @geoip_type = case geoip_initialize.database_type
     when GeoIP::GEOIP_CITY_EDITION_REV0, GeoIP::GEOIP_CITY_EDITION_REV1
@@ -215,6 +215,6 @@ class LogStash::Filters::GeoIP < LogStash::Filters::Base
     # Unfortunately, :pread requires the io-extra gem, with C extensions that
     # aren't supported on JRuby. If / when :pread becomes available, we can stop
     # needing thread-local access.
-    Thread.current[threadkey] ||= ::GeoIP.new(@database)
+    Thread.current[threadkey] ||= ::GeoIP.new(@database, { :preload => true })
   end
 end # class LogStash::Filters::GeoIP


ebuildy:
With data preloading, I think you don't need the Thread stuff, isnit?

suyograo:
@ebuildy can we close this in favor of https://github.com/logstash-plugins/logstash-filter-geoip/pull/64 ? which'll use the new API?

ebuildy:
Absolutely!

I have worked on the  future version, dont need  this any more.

Thanks,

UnitedMarsupials:
Does not "preload" mean, you have to restart your entire Logstash process every time you update the GeoIP-database? A restart takes a good 10 seconds here, during which some of the log-sources will be dropped :(

ebuildy:
If you care about data loss, that mean you have setup a HA environment with multiple logstash instances?

So like any other web softwares (Php, nginx, elasticsearch etc...), use the roll pattern, restart each node 1 by 1.

UnitedMarsupials:
I'll take your answer as a "yes", a restart becomes required for changes to GeoIP-database to take effect...
Your points are fair, and we are working on making our setup HA. However, the present Logstash instance has been running for over 6 months non-stop. Implementing a periodic update of the database would necessitate much more frequent (weekly? daily?) restarts, than what we are currently used to :-(

robcowart:
@UnitedMarsupials my standard recommendation for production environments is always...

logstash_inputs -> message queue -> logstash_processing -> elasticsearch

where...

logstash_inputs is very basic, catch the data using whichever input plugins you require (this is basically you current `input {}` section) and output to a message queue

redis is extremely easy to setup (and fast) as a message queue which holds data temporarily until it is picked up by logstash_processing.

logstash_processing pulls data off from the queue, does whatever processing you need and output to elasticsearch (this would be mostly what is in your current `filter {}` and `output {}` sections).

There are two benefits to this approach. 1. Data volume peaks can be buffered in the message queue in situations where throughput of logstash processing and elasticsearch indexing would otherwise be exceeded (NOTE: persistent queues in current version of Logstash provide this same benefit); 2. You can restart the logstash_processing instance as needed to update the configuration, e.g. a new GeoIP DB, and incoming data will still be received and wait in the queue until logstash_processing has recovered. While there may be a slight data delay during the restart, no data will be lost.

UnitedMarsupials:
Thanks, @robcowart , for the suggestion, but does not it just move the "critical" point around? I mean, now the "logstash_inputs" part can not be restarted without data-loss, can it?.. And we've just added the message-queue to the mix, which needs to be made HA as well...
May as well bite the bullet and make a properly HA Logstash -- or augment the `geoip`-plugin to automatically reload the database as needed :)

robcowart:
To some degree you are correct. However the logstash_inputs is so simple that it would almost never need to be restarted, and because it config is so small it can restart much faster than a larger more complex pipeline. Here is an example of such a pipeline that outputs to redis.

```
input {
    beats {
        type => "beats"
        port => "5044"
        client_inactivity_timeout => 180
    }
    udp {
        type => "ipfix"
        port => "4739"
        codec => netflow {
            versions => [10]
            target => "ipfix"
        }
    }
    udp {
        type => "netflow"
        port => "2055"
        codec => netflow {
            versions => [5,9]
        }
    }
    udp {
        type => "netflow"
        port => "9996"
        codec => netflow {
            versions => [5,9]
        }
    }
    udp {
        type => "sflow"
        port => "6343"
        codec => sflow { }
    }
    udp {
        type => "syslog"
        port => "514"
    }
    udp {
        type => "syslog"
        port => "15514"
    }
    snmptrap {
        type => "snmptrap"
        port => "162"
    }
}
output {
    redis {
        host => "XXXXXX"
        port => "6379"
        password => "XXXXXX"
        data_type => "list"
        key => "logstash:mqi"
    }
}
````
Every system will have its weak points and its strong points, it is just a question of what compromises you are willing to accept.

ebuildy:
My 2 cents, (500k users, 20millions events per day), I use RabbitMQ in front of Logstash, and Php app. send directly to RabbitMQ (no logstash input nodes).

PHP --> RabbitMQ --> logstash --> elasticsearch (--> Spark --> HDFS)

jordansissel:
There seems to be concern about how to update the GeoIP database file without restarting Logstash -- such a feature is fine with me, so please open a new issue about this.

PR #64, opened 2016-03-19
Upgrade to Geoip 2.5.0

garyelephant:
Upgrade to Geoip 2.5.0


vendor.json
```diff
@@ -1,10 +1,6 @@
 [
     {
-        "url": "http://logstash.objects.dreamhost.com/maxmind/GeoLiteCity-2013-01-18.dat.gz",
-        "sha1": "15aab9a90ff90c4784b2c48331014d242b86bf82"
-    },
-    {
-        "url": "http://logstash.objects.dreamhost.com/maxmind/GeoIPASNum-2014-02-12.dat.gz",
-        "sha1": "6f33ca0b31e5f233e36d1f66fbeae36909b58f91"
+        "url": "http://geolite.maxmind.com/download/geoip/database/GeoLite2-City.mmdb.gz",
```
ebuildy: Here I think you should use logstash CDN, how?

ebuildy:
Great job! 

Thanks you

suyograo:
nice work @garyelephant .. we'll review this soon.

suyograo:
@garyelephant I took an initial look, and we definitely don't want to upload jars to the repo. They are only needed for build time, so we should use https://github.com/mkristian/jar-dependencies. I'll add an example here.

vendor.json
suyograo: maxmind should be fine.

ebuildy:
Great tip!

@garyelephant I am on it, I am doing the PR on your repo. in a few minutes.

Cheers,

garyelephant:
@suyograo ,  @ebuildy Thanks, please show me how to use jar jar-dependencies, I have been exhausted in resolving jar dependency problem for days.

ebuildy:
Lost some hours because https://github.com/mkristian/jar-dependencies/pull/40 ;-( 

I will exclude only google.http, since logstash 2.2 doesnot bring fasterxml library yet, but this is something we have to watch for the next version (or is there any way to version plugin according logstash version?)

I am doing the PR right now,

Cheers,

ebuildy:
PR here ==> https://github.com/garyelephant/logstash-filter-geoip/pull/2

Thanks,

suyograo:
@ebuildy fantastic, I was gonna look into this on Monday :) Thanks for the PR. I've left a comment there, just want to make sure all tests pass with jar-dependencies

talevy:
awesome! I will also take a look at this tomorrow!

talevy:
This is looking good!

to run it on my mac, I ran the following

```
bundle install
rake vendor
rake install_jars
bundle exec rspec
```
(tests pass)

do you think we need to include all of the makefile and dockerfile logic within this repo? It is great for testing within a container, but we have decided that we do not wish to make plugin tests and builds depend on things like Docker.

lib/logstash/filters/geoip.rb
```diff
@@ -82,139 +86,121 @@ class LogStash::Filters::GeoIP < LogStash::Filters::Base
...
+    @logger.info("Using geoip database", :path => @database)
 
-    # This is wrapped in a mutex to make sure the initialization behavior of LOOKUP_CACHES (see def above) doesn't create a dupe
-    LOOKUP_CACHE_INIT_MUTEX.synchronize do
-      self.lookup_cache = LOOKUP_CACHES[@geoip_type] ||= LruRedux::ThreadSafeCache.new(1000)
+    db_file = JavaIO::File.new(@database)
+    begin
+      @parser = DatabaseReader::Builder.new(db_file).withCache(CHMCache.new(@cache_size)).build();
+    rescue Java::ComMaxmindDb::InvalidDatabaseException => e
+      # do nothing
```
talevy: we should expose this exception to the user when it occurs

lib/logstash/filters/geoip.rb
```diff
@@ -51,8 +55,8 @@ class LogStash::Filters::GeoIP < LogStash::Filters::Base
   # are included in the event.
   #
   # For the built-in GeoLiteCity database, the following are available:
-  # `city_name`, `continent_code`, `country_code2`, `country_code3`, `country_name`,
-  # `dma_code`, `ip`, `latitude`, `longitude`, `postal_code`, `region_name` and `timezone`.
+  # `city\_name`, `continent\_code`, `country\_code2`, `country\_code3`, `country\_name`,
+  # `dma\_code`, `ip`, `latitude`, `longitude`, `postal\_code`, `region\_name` and `timezone`.
   config :fields, :validate => :array
```
talevy: maybe we should default this to an empty list

```ruby
config :fields, :validate => :array, :default => []
```

lib/logstash/filters/geoip.rb
```diff
@@ -82,139 +86,121 @@ class LogStash::Filters::GeoIP < LogStash::Filters::Base
...
+      response = @parser.city(ip_address)
+      country = response.getCountry()
+      subdivision = response.getMostSpecificSubdivision()
+      city = response.getCity()
+      postal = response.getPostal()
+      location = response.getLocation()
+
+      geo_data_hash = Hash.new()
+
+      if @fields.nil? || @fields.empty? || @fields.include?("city_name")
```
talevy: instead of re-doing all these checks, here is a commit showing a potential refactor I think would make things more readable

https://github.com/logstash-plugins/logstash-filter-geoip/commit/55abe223aa3e9fe726527629e177dda10f990abe

lib/logstash/filters/geoip.rb
talevy: actually, ignore this. I would prefer: https://github.com/logstash-plugins/logstash-filter-geoip/pull/64/files#r57260455

lib/logstash/filters/geoip.rb
garyelephant: @talevy When I add `logger.error` to line 107, it cannot pass spec test, how to solve this ?

The code I added is 
```
@logger.error("Is the Geoip2 MMDB database invalid or corrupted ?", :exception => e, :field => @source)
```

The Error bundle exec rspec output:
```
Failures:
  1) LogStash::Filters::GeoIP an invalid database should return the correct sourcefield in the logging message "{"ip":"8.8.8.8"}" when processed
     Failure/Error: expect(LogStash::Filters::GeoIP.logger).to receive(:error).with(anything, include(:field => "ip"))
       (#<Cabin::Channel:0x7d2b82ef>).error(anything, include {:field => "ip"})
           expected: 1 time with arguments: (anything, include {:field => "ip"})
           received: 2 times with arguments: (#<RSpec::Mocks::ArgumentMatchers::AnyArgMatcher:0x5af6ae34>, #<RSpec::Matchers::BuiltIn::Include:0x79270693 @actual={:exception=
>#<NoMethodError: undefined method `city' for nil:NilClass>, :field=>"ip", :event=>#<LogStash::Event:0x240c243e @metadata={}, @accessors=#<LogStash::Util::Accessors:0x191f74e
f @store={"ip"=>"8.8.8.8", "@version"=>"1", "@timestamp"=>"2016-03-24T02:19:15.058Z", "geoip"=>{}}, @lut={"ip"=>[{"ip"=>"8.8.8.8", "@version"=>"1", "@timestamp"=>"2016-03-24T
02:19:15.058Z", "geoip"=>{}}, "ip"], "geoip"=>[{"ip"=>"8.8.8.8", "@version"=>"1", "@timestamp"=>"2016-03-24T02:19:15.058Z", "geoip"=>{}}, "geoip"]}>, @data={"ip"=>"8.8.8.8",
"@version"=>"1", "@timestamp"=>"2016-03-24T02:19:15.058Z", "geoip"=>{}}, @metadata_accessors=#<LogStash::Util::Accessors:0x29b1b28c @store={}, @lut={}>, @cancelled=false>}, @
expected=[{:field=>"ip"}]>)
     # ./spec/filters/geoip_spec.rb:212:in `(root)'
     # /home/nitrous/.rvm/gems/jruby-1.7.19/gems/rspec-wait-0.0.8/lib/rspec/wait.rb:46:in `(root)'
Finished in 1.74 seconds (files took 4.64 seconds to load)
17 examples, 1 failure
Failed examples:
rspec ./spec/filters/geoip_spec.rb:211 # LogStash::Filters::GeoIP an invalid database should return the correct sourcefield in the logging message "{"ip":"8.8.8.8"}" when pro
cessed
```

ebuildy:
@talevy sure we can remove ./dev folder, this is not used by the plugin itself, I use it to develop everywhere, and change logstash version w/o break my laptop.

.dockerignore
```diff
@@ -0,0 +1,2 @@
+./*.gem
```
talevy: can we remove this file?

.gitignore
```diff
@@ -2,3 +2,5 @@
 Gemfile.lock
 .bundle
 vendor
+.idea
+lib/com
```
talevy: we shouldn't have vendor files in `lib/com`, this can be removed. `.idea` is fine

lib/logstash/filters/geoip.rb
```diff
@@ -82,139 +90,98 @@ class LogStash::Filters::GeoIP < LogStash::Filters::Base
   # of the same geoip_type share the same cache. The last declared cache size will 'win'. The reason for this is that there would be no benefit
   # to having multiple caches for different instances at different points in the pipeline, that would just increase the
   # number of cache misses and waste memory.
-  config :lru_cache_size, :validate => :number, :default => 1000
+  config :cache_size, :validate => :number, :default => 1000
```
talevy: for backwards compatibility, can we leave `lru_cache_size` to work? and maybe call it deprecated

lib/logstash/filters/geoip.rb
talevy: the test just needs to be updated to make this pass and expect that logger call

lib/logstash/filters/geoip.rb
```diff
@@ -82,139 +90,98 @@ class LogStash::Filters::GeoIP < LogStash::Filters::Base
...
+      response = @parser.city(ip_address)
+      country = response.getCountry()
+      subdivision = response.getMostSpecificSubdivision()
+      city = response.getCity()
+      postal = response.getPostal()
+      location = response.getLocation()
+
+      geo_data_hash = Hash.new()
+
+      @fields.each do |field|
```
talevy: nice!

.gitignore
suyograo: @talevy `rake install_jars` (from jar-dependencies) brings in jars to `lib/com`. We may have to change `JARS_HOME` to `vendor`

lib/logstash/filters/geoip.rb
suyograo: +1 to make it deprecated.

CONTRIBUTORS
```diff
@@ -17,6 +17,7 @@ Contributors:
 * Vincent Batts (vbatts)
 * avleen
 * Guy Boertje (guyboertje)
+* Thomas Decaux (qwant)
```
suyograo: @garyelephant Add yourself to this list

lib/logstash/filters/geoip.rb
```diff
@@ -51,9 +55,13 @@ class LogStash::Filters::GeoIP < LogStash::Filters::Base
   # are included in the event.
   #
   # For the built-in GeoLiteCity database, the following are available:
-  # `city_name`, `continent_code`, `country_code2`, `country_code3`, `country_name`,
-  # `dma_code`, `ip`, `latitude`, `longitude`, `postal_code`, `region_name` and `timezone`.
-  config :fields, :validate => :array
+  # `city\_name`, `continent\_code`, `country\_code2`, `country\_code3`, `country\_name`,
```
suyograo: Can you remove `\` from `city\_name` etc. We don't need them

.gitignore
talevy: yeah, we should set that up in a vendor dir

lib/logstash/filters/geoip.rb
garyelephant: Could you please show me how to write the test code ?

.dockerignore
garyelephant: @ebuildy @talevy  I will Remove .dockerignore and ./dev

suyograo:
I am little confused by how the jar-dependencies work here at 1) dev time 2) packaging/install time. @joekiller since you worked with kafka jar-dependencies could you please shed some light here? Can you check if the strategy we're using here would work? Basically we'd like to vendor the jars under `LS_HOME/vendor/` when installing the gem

lib/logstash/filters/geoip.rb
talevy: after trying this out, I notice that I receive the same error. This is received because the plugin continues in trying to filter, even thought it knows that the database file is invalid. I think the plugin should fail early and raise an exception, instead of just logging it.

here is what I am thinking:

The change in `register` to log and raise an exception
```ruby
@@ -124,7 +124,8 @@ class LogStash::Filters::GeoIP < LogStash::Filters::Base
     begin
       @parser = DatabaseReader::Builder.new(db_file).withCache(CHMCache.new(@cache_size)).build();
     rescue Java::ComMaxmindDb::InvalidDatabaseException => e
-      # do nothing
+      @logger.error("The Geoip2 MMDB database provided is invalid or corrupt.", :exception => e, :field => @source)
+      raise e
     end
   end # def register
```

The update to the spec to expect this failure to occur
```ruby
@@ -210,9 +210,9 @@ describe LogStash::Filters::GeoIP do
     context "should return the correct sourcefield in the logging message" do
       sample("ip" => "8.8.8.8") do
         expect(LogStash::Filters::GeoIP.logger).to receive(:error).with(anything, include(:field => "ip"))
-        subject
+        expect { subject }.to raise_error
       end
     end
   end
```

garyelephant: @talevy I finished this.

talevy:
I believe there are a few things left to do before we should go ahead and merge this

1. rebase to resolve conflicts

2. I am getting gem install issues on logstash 2.2.2 (looks to be installed successfully)
```
logstash-2.2.2/bin/plugin install logstash-filter-geoip/logstash-filter-geoip-2.0.5-java.gem 
Validating logstash-filter-geoip/logstash-filter-geoip-2.0.5-java.gem
Installing logstash-filter-geoip
Error Bundler::InstallError, retrying 1/10
An error occurred while installing logstash-filter-geoip (2.0.5), and Bundler cannot continue.
Make sure that `gem install logstash-filter-geoip -v '2.0.5'` succeeds before bundling.
Error Bundler::InstallError, retrying 2/10
```
are you seeing the same?

3. we should place our jars in a consistent place besides the `lib` directory. To do this we can update the rake command for `install_jars` like this:
```
task :install_jars do
  ENV['JARS_HOME'] = Dir.pwd + "/vendor/jar-dependencies/runtime-jars"
  Jars::Installer.vendor_jars!
end
```

let me know what you think!

.gitignore
talevy: once the change to `JARS_HOME` is made, we can go ahead and remove this line from this file

suyograo:
@garyelephant also can you bump the version in gemspec to 3.0.0. This may not be fully compatible.

ebuildy:
@talevy not sure JAR_HOME works, I believe the library take the require_path:

https://github.com/mkristian/jar-dependencies/blob/master/lib/jars/installer.rb#L199

I search a logstash plugin that could teach me the best practice for that, but didn't find any official plugin yet, any suggestion?

Thanks you,

talevy:
interesting, works for me. I am not too familiar with jar-dependencies either, this environment variable seems to be a supported solution though.

garyelephant:
@talevy @ebuildy All things are done and I have never seen gem install issues on logstash 2.2.2.

garyelephant:
One more things:

1.  When I start logstah with this new logstash-filter-geoip installed, it will print such line of log:

```
file:/home/nitrous/softwares/logstash-2.2.2/vendor/jruby/lib/jruby.jar!/jruby/java/core_ext/module.rb:38 warning: replacing File with Java::JavaIo::File in constant 'File on
class/module JavaIO
```

Should we hide it ?

You can test it with this conf

```
# test.conf
input {
  stdin {}
}
filter {
  geoip {
    source => "message"
  }
}
output {
  stdout { codec=>rubydebug }
}
```

```
$ ./bin/logstash -f test.conf
# input: 8.8.8.8
```

talevy:
thanks @garyelephant! Yes, it would be great if that warning were hidden from the user. It is the decision of the plugin, not the user.

garyelephant:
I don't know why the travis build failed, it succeeded in my laptop.

garyelephant:
@talevy Any updates ?

lib/logstash-filter-geoip_jars.rb
```diff
@@ -0,0 +1,8 @@
+# this is a generated file, to avoid over-writing it just delete this comment
```
joekiller: This file should be removed and not checked in as it is generated automatically when the vendor_jars command is run.

logstash-filter-geoip.gemspec
```diff
@@ -22,8 +23,9 @@ Gem::Specification.new do |s|
   # Gem dependencies
   s.add_runtime_dependency "logstash-core-plugin-api", "~> 1.0"
 
-  s.add_runtime_dependency 'geoip', ['>= 1.3.2']
-  s.add_runtime_dependency 'lru_redux', "~> 1.1.0"
+  s.requirements << "jar com.maxmind.geoip2:geoip2, 2.5.0, :exclusions=> [com.google.http-client:google-http-client]"
+
+  s.add_development_dependency "jar-dependencies"
```
joekiller: you could also add the following so jar-dependencies doesn't wrangle it in when you do vendor_jars

    s.add_development_dependency 'ruby-maven', '~> 3.3'

vendor.json
```diff
@@ -1,10 +1,6 @@
...
+        "sha1": "e965584b9f3044670a7a0651b1a1ea83fef47953"
```
joekiller: sha1 appears to have changed to `cb456d0cc5c629c4e18186243f06b9f6911b427d`

joekiller:
The build is blowing up because the jar's aren't there as they are development depedencies. You need a file in the root called `.travis.yml` with the following:

```
sudo: false
language: ruby
cache:
- bundler
- directories:
  - $HOME/.m2
rvm:
- jruby-19mode
before_install:
- gem install jar-dependencies
- bundle install
- rake install_jars
```

Then it should pass fine.

joekiller:
@suyograo @talevy looks like you got the jar stuff figured out otherwise

joekiller:
fyi I added `- rake vendor` to the above.

.gitignore
joekiller: It isn't very good to ignore vendor or lib/com because when you want to do a clean test, git clean -fd doesn't work correctly.

joekiller: ie, you just don't check those in.

talevy:
this is also going to be introduced, so just adding the `rake install_jars` should be necessary after https://github.com/logstash-plugins/logstash-filter-geoip/pull/68 goes in

garyelephant:
@joekiller There is no `.travis.yml` in my fork, maybe this file can't be forked.

talevy:
@garyelephant do you mind adding a `.travis.yml` file?

travis may have decided to just use master's script. we should add a script to this PR so it'll properly be tested by the CI

garyelephant:
@talevy Please have a look.

joekiller:
+1

ebuildy:
I am a bit lost with jar-dependencies stuff :/

When logstash installs a plugin, it does a ``gem install XXX``, isnit? That executes jar dependencies hook (see at https://github.com/jruby/jruby/blob/master/lib/ruby/stdlib/rubygems/defaults/jruby.rb#L123), so to resume, JARs are download from the user server?

Thanks you,

joekiller:
@ebuildy because there are a number of external dependencies that need to be bundled in the gem, you would do:

    bundle exec rake vendor
    bundle exec rake install_jars
    gem build logstash-filter-geoip.gemspec

That would get everything into the gem. So in practice everything will be contained in the gem which allows for "offline" installs and a ready to go package.

talevy:
we can fix some of the multiple-jar-versions warnings and bump the plugin version in a follow-up PR. I think we had enough changes to this one. looking good! Thanks for keeping up with it all!

LGTM

ebuildy:
@joekiller thanks you for the explanation, but when final user will install plugin (which is a ``gem install XXXX``), this will re-install JARs within /lib directory?

talevy:
@ebuildy not to my knowledge, that is why it is specified as a development dependency: https://github.com/logstash-plugins/logstash-filter-geoip/blob/master/logstash-filter-geoip.gemspec#L28

garyelephant:
@ebuildy Since I add the following code in [rakefile](https://github.com/logstash-plugins/logstash-filter-geoip/pull/64/files#diff-52c976fc38ed2b4e3b1192f8a8e24cffR11), jars will only stay in `vendor` dir. geoip.rb invokes these jars directly from `vendor/jar-dependencies/runtime-jars`

```
ENV['JARS_HOME'] = Dir.pwd + "/vendor/jar-dependencies/runtime-jars"
ENV['JARS_VENDOR'] = "false"
```

when you do `gem build`, `vendor` dir as well as these jars will be wrapped in gem tar, without changing dir.

ebuildy:
Ok, got the trick! thanks you,

ebuildy:
@garyelephant I have tested this new geoip, run ``rake install_jars`` then build the .gem end run ``plugin install *.gem``, here the error I got from logstash:

> you might need to reinstall the gem which depends on the missing jar or in case there is Jars.lock then resolve the jars with `lock_jars` command\n\nno such file to load -- com/maxmind/db/maxmind-db/1.1.0/maxmind-db-1.1.0 (LoadError)

How the file ``/lib/logstash-filter-geoip_jars.rb`` includes the right files?

````
# this is a generated file, to avoid over-writing it just delete this comment
require 'jar_dependencies'

require_jar( 'com.fasterxml.jackson.core', 'jackson-databind', '2.6.4' )
require_jar( 'com.maxmind.db', 'maxmind-db', '1.1.0' )
require_jar( 'com.maxmind.geoip2', 'geoip2', '2.5.0' )
require_jar( 'com.fasterxml.jackson.core', 'jackson-annotations', '2.6.0' )
require_jar( 'com.fasterxml.jackson.core', 'jackson-core', '2.6.4' )
````

I can see jars are present

````
/opt/logstash/vendor/local_gems/08928a1c/logstash-filter-geoip-2.0.7-java/vendor/jar-dependencies/runtime-jars/geoip2-2.5.0.jar
````

Whereas this works fine if I run ``rake install_jars`` on the same machine than logstash before create the gem. Did I forget something?

**Comment by joekiller** [comment]
This confusion is why I stuck with just letting it dump the jars into lib
for packaging for jruby-kafka

**Comment by garyelephant** [comment]
@ebuildy So I gave you a wrong explanation

**Comment by garyelephant** [comment]
@suyograo @talevy @joekiller .

For now, I only know if we remove these two lines from rakefile, this plugin will work fine in installation and runtime.But this will result in installing jars in `lib/com`. 
```
ENV['JARS_HOME'] = Dir.pwd + "/vendor/jar-dependencies/runtime-jars"
ENV['JARS_VENDOR'] = "false"
```

Otherwise, it will promote this(I've tested it in logstash 2.2.2, 2.3.1):

```
The error reported is:
  Couldn't find any filter plugin named 'geoip'. Are you sure this is correct? Trying to load the geoip filter plugin resulted in this error: no such file to load -- logstash
-filter-geoip_jars
```

Any better idea ?

**Comment by ebuildy** [comment]
For now the  author of jar depencies module told me there is no  way yet
to put jars outside lib folder and include  it.

**Comment by garyelephant** [comment]
@suyograo @talevy The version was bumped to `3.0.0-beta`, but these lines of code are still in `Rakefile`. Haven't users of this plugin see any jar dependency problem ? I believe these code should be removed and let jars to be installed in `lib` dir.

```
ENV['JARS_HOME'] = Dir.pwd + "/vendor/jar-dependencies/runtime-jars"
ENV['JARS_VENDOR'] = "false"
```

talevy:
@garyelephant seemed to have work for me. I will verify once again and update you

talevy:
@garyelephant looks like my environment wasn't clean. you are correct, my recommendation was incorrect. I will update the repo before we release. thanks!

talevy:
@garyelephant @ebuildy 

https://github.com/logstash-plugins/logstash-filter-geoip/pull/71 should fix the issue.

May make sense to move to the new way of doing things in jar-dependencies with this fix: https://github.com/mkristian/jar-dependencies/issues/44, but that is not released yet.

suyograo:
@garyelephant @ebuildy many thanks for your contribution here. We'll be releasing this plugin soon with 5.0.0-alpha2 and mention you in the blogs. Also a shout out here: https://www.elastic.co/blog/logstash-lines-2016-05-02

lifeofguenter:
I am assuming this won't be added into 2.x?

ebuildy:
Nop like GOT, you must wait :-( if you want it, install
logstash-filter-geoip2 , this is the same plugin (dont forget the
configuration geoip2)
Le 6 juil. 2016 2:34 PM, "Gunter Grodotzki" <notifications@github.com> a
écrit :

PR #73, opened 2016-04-28
Use empty result when location is nil

### LINKED ISSUE #70: v3.0.0-beta Sometimes returns nil value for location field
Testing new version with Maxmind geoIp2 Java, for some IP adresses:

- 113.208.89.219
- 67.209.98.194
- 217.15.165.185

The plugins returns a ``nil`` values for all fields, which is OK because elasticsearch will cast nil as empty string, but in the case of ``location`` field, if you setup ``geo`` mapping, ES will raise a ``mapping_exception`` error, hence document is not indexed.

I know a solution is to use https://www.elastic.co/guide/en/elasticsearch/reference/current/ignore-malformed.html. But I think the plugin should not include ``location`` field if it contains nil values.

suyograo:
For some IPs location and most fields are nil. Its better to return empty geoip than Elasticsearch barfing with mapping errors.

Also added tests for IPv6 now that GeoIP2 supports it
Fixes #70

### DIFF
--- a/lib/logstash/filters/geoip.rb
+++ b/lib/logstash/filters/geoip.rb
@@ -153,25 +153,14 @@ class LogStash::Filters::GeoIP < LogStash::Filters::Base
     begin
       ip = event[@source]
       ip = ip.first if ip.is_a? Array
+      geo_data_hash = Hash.new
       ip_address = InetAddress.getByName(ip)
       response = @parser.city(ip_address)
-      
-      if response.nil?
-        tag_unsuccessful_lookup(event)
-        set_data(event)
-        return
-      else  
-        geo_data_hash = populate_data(event, response, ip_address)
-      end
-
+      populate_geo_data(response, ip_address, geo_data_hash)
     rescue com.maxmind.geoip2.exception.AddressNotFoundException => e
       @logger.debug("IP not found!", :exception => e, :field => @source, :event => event)
-      set_data(event)
-      return
     rescue java.net.UnknownHostException => e
       @logger.error("IP Field contained invalid IP address or hostname", :exception => e, :field => @source, :event => event)
-      set_data(event)
-      return
     rescue Exception => e
       @logger.error("Unknown error while looking up GeoIP data", :exception => e, :field => @source, :event => event)
       # Dont' swallow this, bubble up for unknown issue
@@ -180,25 +169,27 @@ class LogStash::Filters::GeoIP < LogStash::Filters::Base
 
     event[@target] = geo_data_hash
 
+    if geo_data_hash.empty?
+      tag_unsuccessful_lookup(event)
+      return
+    end
+
     filter_matched(event)
   end # def filter
   
-  def populate_data(event, response, ip_address)
+  def populate_geo_data(response, ip_address, geo_data_hash)
     country = response.getCountry()
     subdivision = response.getMostSpecificSubdivision()
     city = response.getCity()
     postal = response.getPostal()
     location = response.getLocation()
 
-    geo_data_hash = Hash.new()
-    
     # if location is empty, there is no point populating geo data
     # and most likely all other fields are empty as well
     if location.getLatitude().nil? && location.getLongitude().nil?
-      tag_unsuccessful_lookup(event)
-      return geo_data_hash
+      return
     end
-    
+
     @fields.each do |field|
       case field
       when "city_name"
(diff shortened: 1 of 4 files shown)


lib/logstash/filters/geoip.rb
```diff
@@ -152,67 +155,98 @@ def filter(event)
...
+      
+      if response.nil?
+        tag_unsuccesful_lookup(event)
+        set_data(event)
+        return
+      else  
+        geo_data_hash = populate_data(event, response, ip_address)
       end
-
+
```
talevy: whitespace?

lib/logstash/filters/geoip.rb
```diff
@@ -152,67 +155,98 @@ def filter(event)
...
       return
     rescue Exception => e
       @logger.error("Unknown error while looking up GeoIP data", :exception => e, :field => @source, :event => event)
-      event[@target] = {}
-      return
+      # Dont' swallow this, bubble up for unknown issue
+      raise e
     end
-
+
```
talevy: whitespace?

spec/filters/geoip_spec.rb
```diff
@@ -186,6 +186,36 @@
...
+        let(:ipstring) { "113.208.89.21" }
+
+        it "should set the target field to an empty hash" do
+          expect(event["geoip"]).to eq({})
+          expect(event["tags"]).to include("_geoip_lookup_failure")
+        end
+      end
+      
+      context "when IP is IPv6 format for localhost" do
+        # regression test for issue https://github.com/logstash-plugins/logstash-filter-geoip/issues/51
```
suyograo: copy-paste

lib/logstash/filters/geoip.rb
```diff
@@ -152,67 +155,98 @@ def filter(event)
...
+    return geo_data_hash
+    
+  end
+  
+  def tag_unsuccessful_lookup(event)
+    @logger.debug? && @logger.debug("IP #{event[@source]} was not found in the database", :event => event)
+    @tag_on_failure.each{|tag| event.tag(tag)}
+  end
+  
+  def set_data(event, geo_data = {})
```
talevy: is this method necessary? 

maybe a better approach is to have a `geo_data` object that is instantiated with `Hash.new` and then gets populated by the above `populate_data` method.

lib/logstash/filters/geoip.rb
```diff
@@ -119,6 +119,9 @@ class LogStash::Filters::GeoIP < LogStash::Filters::Base
   # to having multiple caches for different instances at different points in the pipeline, that would just increase the
   # number of cache misses and waste memory.
   config :lru_cache_size, :validate => :number, :default => 1000
+  
+  # Tags the event on failure to look up geo information. This can be used in later analysis.
+  config :tag_on_failure, :validate => :array, :default => ["_geoip_lookup_failure"]
```
talevy: is there a reason we want to make this configurable?

talevy: maybe this should be a boolean?

suyograo: Wanted this to be consistent with grok filter https://github.com/logstash-plugins/logstash-filter-grok/blob/master/lib/logstash/filters/grok.rb#L204

I thought about the array value vs string. I think I understand why this is an array -- you could have additional tags based on hostname where this is running or something like that for further debugging. 

IMO configurable is nice, users can tweak it

lib/logstash/filters/geoip.rb
suyograo: Hmm, i thought I can dry up usages like here: https://github.com/logstash-plugins/logstash-filter-geoip/pull/73/files#diff-0ceaeedc6497c6d61fa5ae1d2db1dc58R173

suyograo: Also makes it easier when Java Event API work needs to be done :)

lib/logstash/filters/geoip.rb
talevy: sounds good

lib/logstash/filters/geoip.rb
```diff
@@ -152,67 +155,97 @@ def filter(event)
...
+      # Dont' swallow this, bubble up for unknown issue
+      raise e
     end
 
     event[@target] = geo_data_hash
 
     filter_matched(event)
   end # def filter
+  
+  def populate_data(event, response, ip_address)
```
talevy: maybe it would be better to mutate `event` outside of this method, using the `geo_data_hash` that is returned as the signal as to what to do

suyograo:
@talevy let me know if this is better

spec/filters/geoip_spec.rb
```diff
@@ -174,7 +174,7 @@
         end
 
         it "should not have added any tags" do
```
talevy: new behavior not in sync with test name?

suyograo: oh crap, let me change

suyograo:
@talevy updated

talevy:
LGTM

PR #77, opened 2016-05-11
Allow include/exclude fields in the geoip filter

### LINKED ISSUE #6: allow the inclusion/exclusion of the "extra" location field
The `location` field is dynamically added to the geoip structure when both `latitude` and `longitude` fields are present https://github.com/logstash-plugins/logstash-filter-geoip/blob/master/lib/logstash/filters/geoip.rb#L141-L144 to provide a GeoJSON compatible fields structure.

The `fields` config https://github.com/logstash-plugins/logstash-filter-geoip/blob/master/lib/logstash/filters/geoip.rb#L48 should also support including/excluding this `location` field to avoid unnecessary duplicate info.

purbon:
* make location the default field, others could be added through the fields config option
* updated spec with detailed test, plus version bump an changelog update

Fixes #6

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,19 +1,22 @@
+## 5.0.0
+  - [#6] [Breaking change] Allow the inclusion/exclusion of the "extra", location  is the default one, but all others could be added through the fields parameter. See https://github.com/logstash-plugins/logstash-filter-geoip/issues/6 for details.
+
 ## 4.0.0
   - Update the plugin to the version 2.0 of the plugin api, this change is required for Logstash 5.0 compatibility. See https://github.com/elastic/logstash/issues/5141
   - GA release for GeoIP2 database, compatible with LS 5.x
 
-# 3.0.0-beta3
+## 3.0.0-beta3
  - Return empty result when IP lookup fails for location field (#70)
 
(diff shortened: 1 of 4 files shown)


purbon:
@colinsurprenant you started #6, so you might be awesome to review this.

purbon:
@ph your thoughts are very welcome here too.

electrical:
👎 This is beyond a breaking change. Its makes no sense to strip out all useful fields by default and only have the location one.
Also the linked issue talks about filtering out the auto generated `location` field and not making it the only default field..

bastelfreak:
I agree with @electrical, please don't merge that, it kills all the useful defaults.

ph:
I tend to agree with @electrical and @bastelfreak, I am more into provide useful defaults and maybe taking all the fields is a sane default. It might no be the best one, but this is what user have expected for a long time.

Is there any reason we cannot apply the following logic?

- No inclusion or exclusion take all data.
- Inclusion set only take specified fields
- Exclusion remove the specified fields

The think that #6 mention is that we have a duplicate of the location data.

purbon:
> The think that #6 mention is that we have a duplicate of the location data.

then I completely miss understood the meaning of #6

PR #78, opened 2016-05-11
Release 4.0.0 for LS 5.x support and GeoIP2

suyograo:
This adds support for both LS 5.x (new event API) as well as
GeoIP2 database.


ph:
I bless this with a LGTM

PR #79, opened 2016-05-24
set spec encoding to utf-8

talevy:
Test fails because encoding is ASCII

Fixes https://github.com/elastic/logstash/issues/5347


guyboertje:
LGTM

PR #83, opened 2016-06-21
Add metrics info

suyograo:
```
 {
        "id" : "geoip_6346ed52-2773-4b0a-8783-1b0c794fd0e5",
        "cache_size" : 1000,
        "events" : {
          "duration_in_millis" : 67,
          "in" : 1,
          "out" : 1
        },
        "matches" : 1,
        "name" : "geoip"
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 4.0.1
+ - Add metrics for measuring number of matches, failures and other errors.
+
 ## 4.0.0
   - Update the plugin to the version 2.0 of the plugin api, this change is required for Logstash 5.0 compatibility. See https://github.com/elastic/logstash/issues/5141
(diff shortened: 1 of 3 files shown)


ph:
The test failling is similar to https://github.com/elastic/logstash/issues/5518
I have to do more check.

ph:
What I know so far:
- We should always metric available, by default the lazy set will use a `NullMetric`
- No threads are involved in the code.

ph:
Cleaning the *.lock and running with the latest released core allow me to reproduce the error locally.

ph:
Ok, I understand why this is happening this is actually a bug in core which will require a new snapshot release. Its because the NullMetric in the plugin.rb base class is not correctly setup when no pipeline is involved.

ph:
This should be fixed, but it probably require a new snapshot release of the core for the test to pass?

PR #88, opened 2016-07-15
fix "ip not found" spec using 0.0.0.0 instead of random ip

jsvd:
the "ip not found" spec was using a specific ip that now shows up in the latest version of the geo ip database.
this pr changes the test to use 0.0.0.0, which we know won't show up.


purbon:
Good change!! definitely ` 0.0.0.0` will not be in the database.

LGTM

I guess there is no other way to ask for missing IP than ask for either localhost or this, so change looks good! 🚢 :shipit: 🚀

Issue #29, opened 2015-06-06
Option to turn off logging for invalid IP message


jhitze:
What do you all think of having the "IP Field contained invalid IP address or hostname" message be able to turn off?

Like, I know it's going to happen in my IIS logs, as it will put a `-` field in there if the event doesn't have an IP. 

The reason I ask, is because my logstash logs are at 25gb in just a week, mostly because of this log entry.

```shell
grep "IP Field contained invalid IP address or hostname" * | wc -l
375898
```
And that count doesn't even count the number of log lines associated with each of those entries! :(

I'm thinking that it should default to true, so that anyone trying it out for the first time would get the error message like today. Then allow it to be turned off in the config.

```ruby
geoip {
  source => "ip"
  invalid_ip_logging => false
}
```

splashx:
This error also happens if the logstash instance is running in an environment where there is no possibility to resolve the external hostnames.

ricardojdsilva87:
Hello,

I'm having the same problem, but my server has connection to the internet so I can resolve the hostnames on the apache logs.

I have some IP/hostnames on the apache logs that look like this:
115031194-171.lexues.net
213-195-147-170.static.ip.netia.com.pl´
ul.cs-cart.com

When passing trough the geoip filter on the logstash.log I can see the error "IP Field contained invalid IP address or hostname". 

Checking on the machine I can't resolve the hostname 115031194-171.lexues.net into an IP address and to a location, but using only lexues.net the call is valid and I get an IP address. The same applies to the other examples.

My question is if there is any way of trying to resolve those names, or at least using the domain name instead. All this without having to try to identify the correct dns name from a long string, for example lexues.net.

Thanks for the help

ipbastola:
I am facing the similar issue. Any help?

ebuildy:
As you can see on sources:

https://github.com/logstash-plugins/logstash-filter-geoip/blob/91d35650f6defe80b9d1f9a53edb34db6868dcd2/lib/logstash/filters/geoip.rb#L160

It's using logstash logger, 2 solutions:
- Comment this line ^^
- Disable logstash logging

eoconnor:
Another option would be to only call the geoip filter if the field value matches an IP regular expression:

`
if [my-ip-field] =~ "[0-9]*.[0-9]*.[0-9]*.[0-9]*" {
    geoip {
        source => "my-ip-field"
    }
}
`

jordansissel:
[...]

PR #97, opened 2016-10-24
#29 Option to turn off logging for invalid IP message

ebuildy:


jordansissel:
With Logstash 5.0 using log4j2 for logging, and each plugin getting a separate named logger, I feel the best solution here is to use that mechanism (log4j2 configuration) to control Logstash's logging of these kinds of things.

kaisecheng:
This is a very old patch for an aged version, hence close it.

PR #101, opened 2016-11-30
Merge, rather than overwrite target fields.

### LINKED ISSUE #98: Target object gets over-written in v5.0
Fields which, before the event is handled by the geoip filter, were part of the object used as "target" in geoip are missing afterward. In the case below, "ip".

Config:

```
input { stdin { } }

filter {
  csv {
    columns => ["[source][ip]", "[target][ip]"]
     separator => ","
  }
  #-------------
  # geo lookup
  #-------------
  if [source][ip] and [source][ip] != "" {
    geoip {
      source => "[source][ip]"
      fields => [ "city_name", "region_name", "country_name", "country_code2", "location" ]
      target => "source"
    }
  }
  if [target][ip] and [target][ip] != "" {
    geoip {
      source => "[target][ip]"
      fields => [ "city_name", "region_name", "country_name", "country_code2", "location" ]
      target => "target"
    }
  }
}
output { stdout {codec => rubydebug {metadata => true}}}
```

Data:

```
173.9.34.108,173.9.34.107
```

Output from 2.4:

```
{
       "message" => "173.9.34.108,173.9.34.107",
      "@version" => "1",
    "@timestamp" => "2016-11-09T00:54:21.342Z",
          "host" => "crowley",
        "source" => {
                   "ip" => "173.9.34.108",
        "country_code2" => "US",
         "country_name" => "United States",
          "region_name" => "MA",
            "city_name" => "Newton Upper Falls",
             "location" => [
            [0] -71.2187,
            [1] 42.31270000000001
        ]
    },
        "target" => {
                   "ip" => "173.9.34.107",
        "country_code2" => "US",
         "country_name" => "United States",
          "region_name" => "MA",
            "city_name" => "Newton Upper Falls",
             "location" => [
            [0] -71.2187,
            [1] 42.31270000000001
        ]
    }
}
```

Output from 5.0:

```
{
    "@timestamp" => 2016-11-09T00:56:03.029Z,
      "@version" => "1",
          "host" => "crowley",
        "source" => {
            "city_name" => "Mendon",
         "country_name" => "United States",
        "country_code2" => "US",
          "region_name" => "Massachusetts",
             "location" => [
            [0] -71.5439,
            [1] 42.103
        ]
    },
       "message" => "173.9.34.108,173.9.34.107",
        "target" => {
            "city_name" => "Mendon",
         "country_name" => "United States",
        "country_code2" => "US",
          "region_name" => "Massachusetts",
             "location" => [
            [0] -71.5439,
            [1] 42.103
        ]
    }
}
```

untergeek:
3.0.0 did away with field and sub-field merging.  This is a regression.  This PR fixes that.

fixes #98

### DIFF
--- a/lib/logstash/filters/geoip.rb
+++ b/lib/logstash/filters/geoip.rb
@@ -47,7 +47,7 @@ end
 # map visualization).
 #
 # Note: This product includes GeoLite2 data created by MaxMind, available from
-# http://www.maxmind.com. This database is licensed under 
+# http://www.maxmind.com. This database is licensed under
 # http://creativecommons.org/licenses/by-sa/4.0/[Creative Commons Attribution-ShareAlike 4.0 International License]
 
 class LogStash::Filters::GeoIP < LogStash::Filters::Base
@@ -121,7 +121,7 @@ class LogStash::Filters::GeoIP < LogStash::Filters::Base
   # to having multiple caches for different instances at different points in the pipeline, that would just increase the
   # number of cache misses and waste memory.
   config :lru_cache_size, :validate => :number, :default => 1000
-  
+
   # Tags the event on failure to look up geo information. This can be used in later analysis.
   config :tag_on_failure, :validate => :array, :default => ["_geoip_lookup_failure"]
 
@@ -169,16 +169,13 @@ class LogStash::Filters::GeoIP < LogStash::Filters::Base
       raise e
     end
 
-    event.set(@target, geo_data_hash)
-
-    if geo_data_hash.empty?
+    if apply_geodata(geo_data_hash, event)
+      filter_matched(event)
+    else
       tag_unsuccessful_lookup(event)
-      return
     end
-
-    filter_matched(event)
   end # def filter
-  
+
   def populate_geo_data(response, ip_address, geo_data_hash)
     country = response.getCountry()
     subdivision = response.getMostSpecificSubdivision()
@@ -233,6 +230,22 @@ class LogStash::Filters::GeoIP < LogStash::Filters::Base
   def tag_unsuccessful_lookup(event)
     @logger.debug? && @logger.debug("IP #{event.get(@source)} was not found in the database", :event => event)
     @tag_on_failure.each{|tag| event.tag(tag)}
+  end
+
+  def apply_geodata(geo_data_hash, event)
+    # don't do anything more if the lookup result is nil?
+    return false if geo_data_hash.nil?
+    # only do event.set(@target) if the lookup result is not nil
+    event.set(@target, {}) if event.get(@target).nil?
+    # don't do anything more if the lookup result is empty?
+    return false if geo_data_hash.empty?
+    geo_data_hash.each do |key, value|
+      if @fields.include?(key) && value
+        # can't dup numerics
+        event.set("[#{@target}][#{key}]", value.is_a?(Numeric) ? value : value.dup)
+      end
+    end # geo_data_hash.each
+    true
   end
 
 end # class LogStash::Filters::GeoIP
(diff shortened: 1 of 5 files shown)


jordansissel:
Hmm, this test fails:

https://github.com/untergeek/logstash-filter-geoip/blob/8ed594ad0f7432618a9ad7a3c0285d2958d27cfa/spec/filters/geoip_spec.rb#L240-L246

But ... the test is confusing. I don't understand the relationship between "look up an ipv6 address" and "the result should be empty" ? Maybe we remove this test?

jordansissel:
(I don't think your PR changed this test, but since it is failing and I like green travis, I am in favor of removing the test unless we can figure out why it exists)

Otherwise, code review looks good!

untergeek:
Yeah.  The test does not make sense with the change to merge.  I'll remove it.

untergeek:
Actually, I just changed it to accurately test it, and updated the test titles.

lib/logstash/filters/geoip.rb
```diff
@@ -169,16 +169,19 @@ def filter(event)
       raise e
     end
 
-    event.set(@target, geo_data_hash)
+    # event.set(@target, geo_data_hash)
```
jsvd: can we remove this, or is it for documentation purposes?

lib/logstash/filters/geoip.rb
```diff
@@ -169,16 +169,18 @@ def filter(event)
...
-    event.set(@target, geo_data_hash)
+    success = apply_geodata(geo_data_hash, event)
 
     if geo_data_hash.empty?
       tag_unsuccessful_lookup(event)
       return
     end
 
-    filter_matched(event)
+    if success
```
jsvd: should we take care of the else here? if apply_geodata "fails", then there is no tag but also no filter_matched

lib/logstash/filters/geoip.rb
```diff
@@ -235,4 +237,20 @@ def tag_unsuccessful_lookup(event)
     @tag_on_failure.each{|tag| event.tag(tag)}
   end
 
+  def apply_geodata(geo_data_hash, event)
+    # don't do anything more if the lookup result is nil?
```
jsvd: since we check on line 174 if geo_data_hash is empty, should we do that check first and then call populate_geo_data being sure it is not empty/nil?

lib/logstash/filters/geoip.rb
```diff
@@ -235,4 +237,20 @@ def tag_unsuccessful_lookup(event)
...
+    return false if geo_data_hash.nil?
+    # only do event.set(@target) if the lookup result is not nil
+    event.set(@target, {}) if event.get(@target).nil?
+    # don't do anything more if the lookup result is empty?
+    return false if geo_data_hash.empty?
```
jsvd: as stated above, this check is somewhat duplicate with the one at line 174

lib/logstash/filters/geoip.rb
```diff
@@ -235,4 +237,20 @@ def tag_unsuccessful_lookup(event)
...
+    geo_data_hash.each do |key, value|
+      if @fields.include?(key) && !(value.nil?)
```
jsvd: since we're only excluding Numeric from dup, I assume we never expect value to be either `true` or `false`

that means we can actually replace `!(value.nil)` with just `value`

lib/logstash/filters/geoip.rb
untergeek: That's putting the cart before the horse.  It's an interesting pass-by-reference here, as the `populate_geo_data` call is what's populating the `geo_data_hash`.  In other words, it can't be tested for emptiness until after `populate_geo_data` has been called.

jsvd:
LGTM

GlenRSmith:
❤️

Issue #104, opened 2016-12-21
Document GeoIP vs GeoIP2 versioning and compatibility


acchen97:
I've had this come up a few times now and think it's a good idea to make this more clear in our docs.

- Versions >4.0.0 will use legacy MaxMind GeoIP DB, only IPv4 lookups
- Versions 4.0.0+ will use MaxMind GeoIP2 DB, IPv4 and IPv6 lookups

@dedemorton is this maybe something you can help with?

/cc @suyograo

almico:
Related to #92

acchen97:
@dedemorton can we work on this one for the next release?

dedemorton:
@acchen97 Part of the confusion might be because some of the  content is missing from the published doc (basically a code comment is getting pulled into the file instead of the description that's intended). It looks like this right now (but now that we know about the problem, we're fixing it):

https://github.com/logstash-plugins/logstash-filter-geoip/blob/master/lib/logstash/filters/geoip.rb#L34
![image](https://cloud.githubusercontent.com/assets/14206422/22848224/d3d9dbce-efa7-11e6-8c46-b9e3f2e8972b.png)

To see the text that _should_ appear in the doc, look [here](https://github.com/logstash-plugins/logstash-filter-geoip/blob/master/lib/logstash/filters/geoip.rb#L34). 

The doc does mention Maxmind GeoLite2. However, it doesn't mention GeoIP2. When I read the topic as it's written right now, it sounds like the plugin only supports GeoLite2. Do you really mean that this topic should say GeoIP2, or is GeoLite2 correct? 

I think the main change that I need to make to this topic is to add a note saying something like the following:

`NOTE: Plugin versions 4.0.0 and later use the MaxMind GeoLite2 database and support both IPv4 and IPv6 lookups. Versions prior to 4.0.0 use the legacy MaxMind GeoLite database and support IPv4 lookups only.`

Can you review the text and let me know if it sounds OK?

acchen97:
Text LGTM. Thanks @dedemorton

dedemorton:
Closed by https://github.com/logstash-plugins/logstash-filter-geoip/pull/105

PR #105, opened 2017-02-13
Indicate which plugin versions use GeoLite2 vs GeoLite db

dedemorton:
Doc fix for https://github.com/logstash-plugins/logstash-filter-geoip/issues/104


suyograo:
LGTM

PR #108, opened 2017-04-03
Update SHA-1 of the GeoLite2-City database file

joewreschnig:
The previous hash is no longer valid. The new one is for the
GeoLite2-City database from MaxMind which is timestamped March 7th,
2017.

Also add Jars.lock to .gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ Gemfile.lock
 vendor
 .idea
 lib/com
+Jars.lock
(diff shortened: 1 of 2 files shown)


joewreschnig:
I've signed the CLA, but only just. I guess there's an eventually-consistent step somewhere in that process too.

joewreschnig:
Even the Lite data seems to also be on a monthly update schedule, so this is already outdated.

PR #109, opened 2017-04-06
Look up continent/country/region/city GeoName IDs.

joewreschnig:
For backwards-compatibility, GeoName IDs are only included if explicitly
added to the `fields` option.

Also, update the database SHA-1 and fix an unrelated test warning due to
over-broad exception handling.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 *.gem
 Gemfile.lock
+Jars.lock
 .bundle
 vendor
 .idea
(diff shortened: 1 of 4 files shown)


joewreschnig:
If I rewrite this PR for the most recent code, what are the chances it will get looked at before the code is rewritten again?

GeoName IDs are necessary to implement MaxMind's own best-practices: "We strongly discourage you from using a value from any *_name field as a key in a database or map/dict/hash data structure."

suyograo:
@joewreschnig I'll add this soon.

joewreschnig:
Thank you!

joewreschnig:
Any news on this?

Issue #107, opened 2017-03-06
Need more information extracted from GeoIP database not only lat or lng


keefeleen:
We have to extract **province**, **city** information from MaxMind GeoIP2 database when using logstash filter.

But it seems that logstash default geoip plugin can just provide "latitude" and "longitude" info. 
Actually we write a plugin for info extraction, but we strongly recommend official plugin maintainer can update this plugin. Can future version add more fields that can be extracted from GeoIP2 databases? 

Thanks in advance and waiting for your reply.

wiibaa:
Strange report as the default list of field is already quite exhaustive and contains the city_name : https://github.com/logstash-plugins/logstash-filter-geoip/blob/master/lib/logstash/filters/geoip.rb#L67 
Also this is configurable by you to add the desired level of **subdivision** as documented here http://dev.maxmind.com/geoip/geoip2/geoip2-city-country-csv-databases/ 

@keefeleen do you have a restricted **fields** config in your configuration ?

keefeleen:
Thanks for your replay, I checked over again that just as you said the default fields already contains the info we need. Then I found the problem that caused failure when we tried to extract city_name information.

Since GeoIP filter uses GeoIP2-java lib. We found that before reading the database, the DatabaseReader class will compare database's metadata named "database_type" with the data type name we want (eg. "city"), if "database_type" doesn't contain the data type name, it will throw an exception and won't give the IP data we want.

I think the related code is as follows in DatabaseReader.java:

        String databaseType = this.getMetadata().getDatabaseType();
        if (!databaseType.contains(type)) {
            String caller = Thread.currentThread().getStackTrace()[2]
                    .getMethodName();
            throw new UnsupportedOperationException(
                    "Invalid attempt to open a " + databaseType
                            + " database using the " + caller + " method");
        }

Actually we want to know the exact purpose of adding this checking logic, and we also need to know the way to avoid failure when we do not want to use "city" or "country" in "database_type" while reading "city" or "country" information.

Thanks in advance.

keefeleen:
@wiibaa hello, can you take a look at the question above?

Since another team in our company build our own city level GeoIP database which follows MaxMind's standard and provide it for us to use. But the "database_type" the defined doesn't contain "city" in it so we cannot use logstash GeoIP plugin with it.

Is there any way to avoid asking them to rebuild the database and make logstash GeoIP plugin compatible with the existing database?

wiibaa:
[...]

PR #111, opened 2017-04-20
Support ISP database for ASN data and move parts to Java

suyograo:
This adds support to GeoIP2 ISP database which includes the ASN data.

Also, while doing this, it made sense to me to just move most parts to Java since we were already using the java APIs and not the ruby libs. All existing tests pass.


lib/logstash/filters/geoip.rb
```diff
@@ -64,11 +51,7 @@ class LogStash::Filters::GeoIP < LogStash::Filters::Base
   # For the built-in GeoLite2 City database, the following are available:
   # `city_name`, `continent_code`, `country_code2`, `country_code3`, `country_name`,
   # `dma_code`, `ip`, `latitude`, `longitude`, `postal_code`, `region_name` and `timezone`.
-  config :fields, :validate => :array, :default => ['city_name', 'continent_code',
-                                                    'country_code2', 'country_code3', 'country_name',
-                                                    'dma_code', 'ip', 'latitude',
-                                                    'longitude', 'postal_code', 'region_name',
-                                                    'region_code', 'timezone', 'location']
+  config :fields, :validate => :array
```
suyograo: The docs still mention the same defaults, but the actual defaults have moved to Java enum class.

lib/logstash/filters/geoip.rb
```diff
@@ -113,132 +96,44 @@ class LogStash::Filters::GeoIP < LogStash::Filters::Base
   # of the same geoip_type share the same cache. The last declared cache size will 'win'. The reason for this is that there would be no benefit
   # to having multiple caches for different instances at different points in the pipeline, that would just increase the
   # number of cache misses and waste memory.
-  config :lru_cache_size, :validate => :number, :default => 1000
+  config :lru_cache_size, :validate => :number, :default => 1000, :deprecated => "This field has been deprecated in favor of cache_size."
```
suyograo: we were never using this, and it was lingering around

src/main/java/org/logstash/filters/Fields.java
```diff
@@ -0,0 +1,78 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import java.util.Arrays;
+import java.util.EnumSet;
+import java.util.Locale;
+
+enum Fields {
+  ANONYMOUS_SYSTEM_NUMBER("anonymous_system_number"),
```
jordansissel: In ASN, the A is autonomous, not anonymous

lib/logstash/filters/geoip.rb
jordansissel: Not listing a :default here makes the documentation generator (which will change soon with @ph's work) show that the setting has no default value. Could be confusing.

However, I think given @ph/@dede's efforts to change how plugins get documented, this is probably a fine change.

(Mostly noting this for posterity)

src/main/java/org/logstash/filters/GeoIpFilter.java
```diff
@@ -0,0 +1,343 @@
...
+import com.maxmind.geoip2.DatabaseReader;
+import org.logstash.ext.JrubyEventExtLibrary.RubyEvent;
+
+import java.io.File;
+import java.io.IOException;
+import java.net.InetAddress;
+import java.net.UnknownHostException;
+import java.util.*;
+
+public class GeoIpFilter {
```
jordansissel: IP is capitalized, and writing it "Ip" feels wrong. I dont' think we have a firm style guide on this, but my practice has always been acronyms stay all caps. 

This kind of mixing gets confusing when, for example, Java crypto calls it "SSL" but Netty calls it "Ssl" ... 

Do we have a practice for this?

src/main/java/org/logstash/filters/GeoIpFilter.java
```diff
@@ -0,0 +1,343 @@
...
+      throw new IllegalArgumentException("The database provided is invalid or corrupted.", e);
+    } catch (IOException e) {
+      throw new IllegalArgumentException("The database provided was not found in the path", e);
+    }
+    this.desiredFields = createDesiredFields(fields);
+  }
+
+  private Set<Fields> createDesiredFields(List<String> fields) {
+    Set<Fields> desiredFields = EnumSet.noneOf(Fields.class);
+    if (fields == null) {
```
jordansissel: also worth checking if fields.isEmpty() ? Or would we reject an empty fields list at configuration-time?

src/main/java/org/logstash/filters/GeoIpFilter.java
```diff
@@ -0,0 +1,343 @@
...
+    return applyGeoData(geoData, event);
+  }
+
+  private boolean applyGeoData(Map<String, Object> geoData, Event event) {
+    if (geoData == null) {
+      return false;
+    }
+    // only do event.set(@target) if the lookup result is not null
+    if (event.getField(targetField) == null) {
+      event.setField(targetField, Collections.emptyMap());
```
jordansissel: is this necessary? event.set("[foo][bar]", ...) should create any parent fields if it doesn't exist... I think?

src/main/java/org/logstash/filters/GeoIpFilter.java
suyograo: yeah i was debating this. I do like acronyms as capitalized. I'll change it to `GeoIP`

src/main/java/org/logstash/filters/GeoIpFilter.java
suyograo: no it doesn't. I had to put this in, otherwise, tests fail. It was in the old ruby code as well.

suyograo:
@robacj suggestion accepted. I shortened the fields like you mentioned.

@jordansissel updated after your comments.

src/main/java/org/logstash/filters/GeoIPFilter.java
```diff
@@ -0,0 +1,342 @@
...
+import java.io.File;
+import java.io.IOException;
+import java.net.InetAddress;
+import java.net.UnknownHostException;
+import java.util.*;
+
+public class GeoIPFilter {
+  private static Logger logger = LogManager.getLogger();
+  private static final String CITY_DB_TYPE = "GeoLite2-City";
+  private static final String COUNTRY_DB_TYPE = "GeoLite2-Country";
```
oschwald: The commercial databases have the type `GeoIP2-City` and `GeoIP2-Country`. It'd likely be worth supporting those too, especially given that they are mentioned in the [recent blog post](https://www.elastic.co/blog/geoip-in-the-elastic-stack).

suyograo: @oschwald good call. Updated.

src/main/java/org/logstash/filters/GeoIpFilter.java
```diff
@@ -0,0 +1,343 @@
...
+    City city = response.getCity();
+    Location location = response.getLocation();
+    Continent continent = response.getContinent();
+    Postal postal = response.getPostal();
+    Subdivision subdivision = response.getMostSpecificSubdivision();
+    Map<String, Object> geoData = new HashMap<>();
+
+    // if location is empty, there is no point populating geo data
+    // and most likely all other fields are empty as well
+    if (location.getLatitude() == null && location.getLongitude() == null) {
```
joewreschnig: Since you're rewriting this anyway, please don't keep this mistake. [There are GeoIP records with no lat-lon](https://github.com/logstash-plugins/logstash-filter-geoip/issues/107#issuecomment-292182214) that this plugin should be able to handle.

acchen97:
Minor note, changelog is missing for this release.

zcola:
When can I use it?

hackery:
@zcola you can use it now by updating the plugin within your current logstash:

    > logstash-plugin update logstash-filter-geoip
    Updating logstash-filter-geoip
    Updated logstash-filter-geoip 4.0.4 to 4.1.1

zcola:
@hackery 
I upgraded to 4.2.0 no show isp,, here is my compilation script fragment
~~~
  subdivisions        => [ 'array', 'map'],
    traits        =>  'map',
    location            => 'map',
...
my $tree = MaxMind::DB::Writer::Tree->new(
    ip_version    => 4,
    record_size   => 32,
    database_type => 'GeoIP2-ISP',
    languages     => ['en'],
...


        en      => $arr[14],
                            zh_CN   => $arr[15],
                        },
                    },
                    traits    => {
                            isp   => $arr[17],
                            organization  => $arr[19],
                    },
~~~
logstash return
~~~
{
"@timestamp" => 2017-06-23T08: 01: 36.884Z,
"Message_geo" => {
"City_name" => "Guangzhou",
"Time zone" => "Asia / Shanghai"
"Ip" => "1.0.8.0"
"Latitude" => 23.1167,
"Country_name" => "China"
"Country_code2" => "CN",
"Continent_code" => "AS",
"Country_code3" => "CN",
"Region_name" => "Guangdong",
"Location" => {
"Lon" => 113.25,
"Lat" => 23.1167
},
"Region_code" => "44"
"Longitude" => 113.25
},
"@version" => "1"
"Host" => "xxx"
~~~
Use python to parse the return
```
 }
    },
    "traits": {
        "organization": "\u58a8\u5c14\u672cGoldenit\u6709\u9650\u516c\u53f8",
        "isp": ""
    },
    "subdivisions": [
        {
            "iso_code": "VIC",
            "names": {
                "en": "Victoria",
                "zh_CN": ""
            }
        }
```
Is my compiler script wrong?

PR #115, opened 2017-05-15
Fix packaging to include the default DB

suyograo:


ph:
Look fine for the other plugins https://github.com/logstash-plugins/logstash-filter-useragent doesn't use a vendor json file anymore.

PR #119, opened 2017-06-16
Update GeoIP2 City database

suyograo:
The tests had to be modified because the latest DB does not populate these fields for
test IP 8.8.8.8


original-brownbear:
@suyograo LGTM :)

suyograo:
thanks @original-brownbear

PR #133, opened 2017-11-13
update GeoLite2 sha

jsvd:


jordansissel:
Geoip data test failing due to some precision changes (City "Worcester" now reported from the test IP being "Salem")

jordansissel:
Otherwise LGTM.

yaauie:
I have validated that the file currently being returned from the listed URI has that sha1; do we have any external validation from the provider that it _should_ be that sha1? I see that they list other files with their MD5's, but don't see a checksum of any kind on their site for this particular URI.

jsvd:
There isn't, I checked the md5 and then computed sha1.

PR #134, opened 2017-12-21
Short circuit empty sources, update DB

jakelandis:


spec/filters/geoip_spec.rb
```diff
@@ -93,8 +93,8 @@
 
       it "should set other subfields of 'target' properly" do
         expect(event.get("target").to_hash.keys.sort).to eq(["city_name", "ip", "region_name"])
-        expect(event.get("[target][city_name]")).to eq("Salem")
-        expect(event.get("[target][region_name]")).to eq("New Hampshire")
+        expect(event.get("[target][city_name]")).to eq("Watertown")
+        expect(event.get("[target][region_name]")).to eq("Massachusetts")
```
jakelandis: This looks like it just due to an update to the DB ... the two locations are kinda close:
image

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 5.0.3
+ - Short circuit empty sources (minor performance improvement)
```
jsvd: I'd suggest instead "Skip lookup operation if source field contains an empty string"

jakelandis: Changed as suggested. Thanks!

src/main/java/org/logstash/filters/GeoIPFilter.java
```diff
@@ -114,17 +114,21 @@ public boolean handleEvent(RubyEvent rubyEvent) {
...
     if (input instanceof List) {
       ip = (String) ((List) input).get(0);
+
     } else if (input instanceof String) {
       ip = (String) input;
     } else {
       throw new IllegalArgumentException("Expected input field value to be String or List type");
     }
 
+    if (ip.trim().isEmpty()){
```
jordansissel: maybe a feature worth shipping upstream into the maxmind geoip library?

PR #142, opened 2018-10-29
Update GeoLite2-City.mmdb.gz SHA1

kcm:
Update GeoLite2-City.mmdb.gz SHA1 to match new MaxMind-provided database


kcm:
#143 rolls this up, closing.

PR #143, opened 2018-10-29
Update fields returned and city name

kcm:
MaxMind adjusted some of their data, so now the test IP that resolved to Watertown, MA, US, now resolves to Salem, MA, US.

Additionally, the list of fields returned has changed in another test case, so list all the fields that are now returned.


kcm:
Note: will require #142 to be merged first so that SHA1s match and the build progresses to the point this PR addresses.

kcm:
@danhermann all set for a second look

danhermann:
Looks good for squash+merge. Thanks, @kcm!

PR #157, opened 2019-08-12
Update of GeoLite2 DB and fixed test

andsel:
Updated the SHA1 for the new file and fixed a test related to IP that moved from one city to next closest.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 6.0.2
+  - Update of GeoLite2 DB
+
 ## 6.0.1
   - Fixed deeplink to Elasticsearch Reference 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## 6.0.2
+  - Update of GeoLite2 DB
```
robbavey: Would you mind adding a link to this issue, as shown below
```suggestion
  - Update of GeoLite2 DB  [#157](https://github.com/logstash-plugins/logstash-filter-geoip/pull/157)
```

andsel: Added the suggestion

PR #156, opened 2019-07-22
Fields is missing region_code

untergeek:
`region_code` is still a valid field to extract, but was not in the docs.

On a separate note, I believe that `country_code3` is no longer 3 characters, but 2 (extracted from the same field as `country_code2`), and perhaps should continue to exist, since users may expect it.

PR #158, opened 2019-08-12
Fields is missing region_code

colinsurprenant:
Reboot of #156 

`region_code` is still a valid field to extract, but was not in the docs.


colinsurprenant:
I guess there is no need to bump version for this as the docs will get updated with this change

colinsurprenant:
@untergeek @robbavey

colinsurprenant:
actually for docs to be updated we need to bump+publish the plugin.

colinsurprenant:
LGTM² always better.

colinsurprenant:
v6.0.3 published.

PR #172, opened 2021-01-11
Management database update

kaisecheng:
logstash-filter-geoip currently shipped with Maxmind CC license databases. As Maxmind change the license type, this PR implements the [idea](https://github.com/elastic/logstash/issues/12560) of automatic database update to let users run with a up-to-date proprietary EULA database and thus be compliant. 

#### Online mode
By default, the plugin runs in online mode which manages the download, unzip, daily check, clean up jobs. The new database should be used on-the-fly and if the plugin failed to check update for 30 days, the pipeline of geoip will stop.  There is an ongoing discussion on OSS/ Basic handling and backward compatibility, so a follow-up PR on these topics is expected
```    
geoip { }
```
#### Offline mode
Users can run in offline mode if they don't want the plugin to do the update job, which means users take ownership to manage the database. Logstash < 7.12 will fallback to offline mode regardless of `database` settings because of the dependency of [Agent#stop_pipeline](https://github.com/elastic/logstash/pull/12582/files#diff-f89cc53b47af9c537c413527d74a897735f8339a72b997ad07a5fa62fc152855R209).
```    
geoip {
  database => "PATH_TO_DATABASE_FILE"
}
```
#### Summary of changes
- fix failed test cases
- upgrade gradle for java 11, fix legacy script and warning
- fix broken database download script
- implement database manager 
  - check update
  - download and unzip
  - manage metadata
  - schedule job
  - clean up old database
  - shut down pipeline
  - check OSS (ongoing discussion)
  - add backward compatibility (ongoing discussion)
#### Related issues
Logstash PR https://github.com/elastic/logstash/pull/12582
Meta https://github.com/elastic/logstash/issues/12560


lib/logstash/filters/geoip/database_manager.rb
```diff
@@ -0,0 +1,265 @@
...
+
+  def initialize(geoip, database_path, database_type)
+    @geoip = geoip
+    @mode = get_mode(database_path)
+    @metadata_path = get_file_path("metadata.csv")
+    @database_type = database_type
+    @database_path = patch_database_path(database_path)
+
+    if @mode == :online
+      raise "logstash-filter-geoip is under elastic Basic License. You are running in open source version, hence the pipeline stops" if LogStash::OSS
```
kaisecheng, line 29: We haven't decided yet to block OSS user completely from running this download manager

lib/logstash/filters/geoip/database_manager.rb
```diff
@@ -0,0 +1,273 @@
...
+  # Make sure the path has usable database
+  def assert_database(database_path)
+    raise "failed to load database #{database_path}" unless org.logstash.filters.GeoIPFilter.validate_database(database_path)
+  end
+
+  def check_age
+    metadata = get_metadata.last
+    timestamp = (metadata)? metadata[Column::UPDATE_AT] : 0
+
+    days_without_update = (Time.now.to_i - timestamp.to_i) / (24 * 60 * 60)
```
andsel, line 190: This date calculation could be simplified as:
```ruby
days_without_update = (Date.today - Time.at(timestamp).to_date).to_i
```

lib/logstash/filters/geoip/database_manager.rb
```diff
@@ -0,0 +1,273 @@
...
+    Zlib::GzipReader.open(zip_path) do |gz|
+      ::File.open(database_path, "wb") do |f|
+        f.print gz.read
+      end
+    end
+    database_path
+  end
+
+  # Make sure the path has usable database
+  def assert_database(database_path)
```
andsel, line 182: It raises an exception if something is not ok, so probably `assert_database!` is best

andsel (changes requested):
Hi @kaisecheng I've left some comments. I have big doubt about the `DatabaseManager` class perhaps has too many resposibilities:
- it has download the file
- manage metadata
- implement the validity check

I could imagine to split in 2 halves one that has the behavior to download and unzip the files, the other that manages the metadata and implements the checks; dividing them by abstraction; one that handle more low level duties and the other that implements the core logic of the checks.

lib/logstash/filters/geoip/database_manager.rb
```diff
@@ -0,0 +1,273 @@
...
+# Once a new database version release, DM downloads it, and GeoIP Filter uses it on-the-fly.
+# If the last update timestamp is 25 days ago, a warning message shows in the log; if it is 30 days ago, the GeoIP Filter should shutdown in order to be compliant.
+# There are online mode and offline mode in DM. `online` is for automatic database update while `offline` is for static database path provided by users or Logstash running in <= 7.11
+module LogStash module Filters module Geoip class DatabaseManager
+  include LogStash::Util::Loggable
+
+  def initialize(geoip, database_path, database_type)
+    @geoip = geoip
+    @mode = get_mode(database_path)
+    @metadata_path = get_file_path("metadata.csv")
```
andsel, line 24: I think all this metadata-related behavior could be separated in a class, for example the methods `setup_medata` and the CSV read/write methods could be moved there, this should also simplify the unit testing part

lib/logstash/filters/geoip.rb
```diff
@@ -90,18 +93,12 @@ class LogStash::Filters::GeoIP < LogStash::Filters::Base
 
   public
   def register
-    if @database.nil?
-      @database = ::Dir.glob(::File.join(::File.expand_path("../../../vendor/", ::File.dirname(__FILE__)),"GeoLite2-#{@default_database_type}.mmdb")).first
+    @database_path = @database
```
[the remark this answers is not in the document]
kaisecheng: at the end, I move back with`@database`. GIving two names seems confusing

build.gradle
```diff
@@ -44,80 +45,74 @@ repositories {
...
+  testImplementation group: "com.maxmind.geoip2", name: "geoip2", version: "2.9.0"
+  testImplementation group: "com.maxmind.db", name: "maxmind-db", version: "1.2.2"
+  testImplementation fileTree(dir: logstashCoreGemPath, include: '**/*.jar')
+
+  geolite2('org.elasticsearch:geolite2-databases:20191119') {
+    transitive = false
+  }
 }
 
-task rubyTest(dependsOn: [rubyBootstrap, build]) << {
```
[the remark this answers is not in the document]
kaisecheng: this piece of code is dead. CI use `bundle exec rspec` since the begining

lib/logstash/filters/geoip/database_manager.rb
kaisecheng: Thanks for the feedback and suggestion. I split it to `database_manager`, `database_metadata`, `download_manager` and an util class. I am considering to cache the metadata instead of touching it multiple times, but it is minor to the overall performance given it runs once per day

lib/logstash/filters/geoip/download_manager.rb
```diff
@@ -0,0 +1,88 @@
...
+    Zlib::GzipReader.open(zip_path) do |gz|
+      ::File.open(database_path, "wb") do |f|
+        f.print gz.read
+      end
+    end
+    database_path
+  end
+
+  # Make sure the path has usable database
+  def assert_database!(database_path)
```
andsel, line 73: nitpick: I would named it something like `validate_database!` since `assert_database` recall me something related to unit test instead of runtime validation of data

andsel (approved):
LGTM, left just a little rename suggestion. Thanks to have split the code it looks like more readable.

kaisecheng:
This PR will split to three
1. fix plugin CI
2. GeoIP database service move to logstash core
3. plugin use logstash core to manage database

PR #175, opened 2021-02-16
Upgrade build script

kaisecheng:
This is a PR split from https://github.com/logstash-plugins/logstash-filter-geoip/pull/172 to fix the broken ci

- upgrade gradle for java 11
- fix download task
- remove legacy warning


andsel (approved):
LGTM

PR #176, opened 2021-02-18
use EULA GeoIP2 Database with auto-update in Elastic license

kaisecheng:
This PR is a split from https://github.com/logstash-plugins/logstash-filter-geoip/pull/172
Additionally, the plugin resolves xpack database manager to run auto-update, otherwise, run the offline CC database


kaisecheng:
jenkins test this please

kaisecheng:
jenkins test this please

kaisecheng:
jenkins test this please

lib/logstash/filters/geoip.rb
```diff
@@ -118,4 +117,30 @@ def tag_unsuccessful_lookup(event)
...
+    execution_context.agent.stop_pipeline(pipeline_id)
+  end
+
+  def close
+    @database_manager.close unless @database_manager.nil?
+  end
+
+  def load_database_manager?
+    begin
+      require_relative "#{LogStash::Environment::LOGSTASH_HOME}/x-pack/lib/filters/geoip/database_manager"
```
kaisecheng, line 138: try to load the database manager from xpack

lib/logstash/filters/geoip.rb
```diff
@@ -118,4 +117,30 @@ def tag_unsuccessful_lookup(event)
...
+      true
+    rescue LoadError => e
+      @logger.info("DatabaseManager is not in classpath", :version => LOGSTASH_VERSION, :exception => e)
```
kaisecheng, line 141: log in `info` for OSS user or old Logstash version user

kaisecheng:
This PR is pending on new docker image logstash-8.0.0-SNAPSHOT

kaisecheng:
jenkins test this please

spec/filters/geoip_online_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+        }
+      }
+      CONFIG
+
+      sample("ip" => "173.9.34.107") do
+        insist { subject.get("geoip") }.include?("ip")
+        expect(::File.exist?(METADATA_PATH)).to be_falsey
+      end
+    end
+  end if MAJOR < 7 || (MAJOR == 7 && MINOR <= 12)
```
andsel, line 64: I've a doubt on this conditional tests. They decision of which is executed is based on the values present in https://github.com/logstash-plugins/.ci/blob/b20255c2da0c7fa82e9a247f6366e2da61b47e6c/travis/matrix.yml#L18.
Should we grant runs of these tests with Logstash `7.12` _and_ Logstash `7.13+`? In this way we grant no regressions in future releases of the plugin till the End-of-Life for `7.12`, then we can drop those. 
WDYT?

andsel (changes requested):
@kaisecheng I'left a couple of comments, one minor method extraction and a doubt about versioned tests.

spec/filters/geoip_online_spec.rb
kaisecheng: As discussed, it is better to pin test against 7.11.0 to prevent losing track on < 7.12 in the future

andsel (approved):
LGTM

PR #177, opened 2021-02-19
[Doc] GeoIP database service

kaisecheng:
This PR explains GeoIP database service auto-update feature

Related issues
https://github.com/elastic/logstash/issues/12560

### DIFF
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -134,9 +134,10 @@ The path to MaxMind's database file that Logstash should use. The default databa
 GeoLite2-City, GeoLite2-Country, GeoLite2-ASN are the free databases from MaxMind that are supported.
 GeoIP2-City, GeoIP2-ISP, GeoIP2-Country are the commercial databases from MaxMind that are supported.
 
-Database auto-update applies to default distribution. When `database` point to user's database path,
-auto-update will be disabled
-{logstash-ref}/plugins-filters-geoip.html#plugins-filters-geoip-database_license[see also database license]
+Database auto-update applies to default distribution. When `database` points to user's database path,
+auto-update will be disabled.
+See
+<<plugins-{type}s-{plugin}-database_license,Database License>> for more information.
 
 [id="plugins-{type}s-{plugin}-default_database_type"]
 ===== `default_database_type`
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -37,6 +37,19 @@ can be https://dev.maxmind.com/geoip/geoip2/geolite2[downloaded from here].
 
 If you would like to get Autonomous System Number(ASN) information, you can use the GeoLite2-ASN database.
 
+[id="plugins-{type}s-{plugin}-database_license"]
+==== Database License
+
+MaxMind changed from releasing the database under a CC license to a proprietary EULA,
+the EULA requires Logstash to update the MaxMind database for our users within 30 days of MaxMind updating their database.
+
+For Logstash default distribution user, the plugin manages database for you.
+With the default setting, an auto-update makes sure database run in the latest version.
+For any reason Logstash fails to download database for 30 days, Logstash will stop the plugin pipeline in order to be compliant.
+If user config `database` option to a path, the plugin does not check for database update, and thus user has to be compliant on their own.
+
+For Logstash open source user, the plugin use the Maxmind CC license database by default.
+
 ==== Details
 
 A `[geoip][location]` field is created if
@@ -116,7 +129,8 @@ GeoLite2-City, GeoLite2-Country, GeoLite2-ASN are the free databases from Maxmin
 GeoIP2-City, GeoIP2-ISP, GeoIP2-Country are the commercial databases from Maxmind that are supported.
 
 If not specified, this will default to the GeoLite2 City database that ships
-with Logstash.
+with Logstash. Database auto-update applies to default distribution.
+{logstash-ref}/plugins-filters-geoip.html#plugins-filters-geoip-database_license[see also database license]
 
 [id="plugins-{type}s-{plugin}-default_database_type"]
 ===== `default_database_type`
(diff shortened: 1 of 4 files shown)


karenzone:
Hi Kaise. You did a nice job describing the licensing changes and behaviors. Thank you.

I'm looking at the `database` option and considering how it impacts the database behavior. It's a bit confusing because we say there's no **default setting**, but yet we talk about **default behavior**. 

I considered something like this: 
```
  * Value type is <<path,path>>
  * There is no default value for this setting.
  * If not specified, the database defaults to the GeoLite2 City database that ships
  with Logstash.
```
Could we add a default setting that preserves the default behavior?  I see it's expecting a path, so that might not work. 

If there's not a more user-friendly solution, perhaps we could just delete the line that says,   "There is no default value for this setting" and add the line "If not specified, the database defaults to the GeoLite2 City database that ships
  with Logstash."
  
 I have some other minor wording suggestions to share, too. But it seems to make sense to resolve this first. Please let me know what you think.

docs/index.asciidoc
```diff
@@ -37,6 +37,19 @@ can be https://dev.maxmind.com/geoip/geoip2/geolite2[downloaded from here].
...
+
+MaxMind changed from releasing the database under a CC license to a proprietary EULA,
+the EULA requires Logstash to update the MaxMind database for our users within 30 days of MaxMind updating their database.
+
+For Logstash default distribution user, the plugin manages database for you.
+With the default setting, an auto-update makes sure database run in the latest version.
+For any reason Logstash fails to download database for 30 days, Logstash will stop the plugin pipeline in order to be compliant.
+If user config `database` option to a path, the plugin does not check for database update, and thus user has to be compliant on their own.
+
+For Logstash open source user, the plugin use the Maxmind CC license database by default.
```
karenzone, line 51:
```suggestion
https://www.maxmind.com[MaxMind] changed from releasing the GeoIP database under
a Creative Commons (CC) license to a proprietary end-user license agreement
(EULA).  The MaxMind EULA requires Logstash to update the MaxMind database
within 30 days of a database update. If Logstash fails to download the database
for 30 days, it will stop the pipeline in order to maintain compliance.

The GeoIP filter plugin can manage the database for users running the Logstash default
distribution, or you can manage
database updates on your own. The behavior is controlled by the `database`
setting. 
When you use the default `database` setting, the auto-update feature ensures that the plugin is
using the latest version of the database.
Otherwise, you are responsible for maintaining compliance.

The Logstash open source distribution uses the MaxMind Creative Commons license
database by default.
```

karenzone (commented):
I added some suggested text inline for consideration and discussion

docs/index.asciidoc
karenzone: Overall: There are some instances of "Maxmind" in some of the older sections. Will you please change them all to  to "MaxMind" for accuracy?

kaisecheng: 👌

kaisecheng:
good catch on the default setting and default behavior. `database` is a path that varies on different systems, so you are right, there is no easy expression to show the default value. So, I took your suggestion.

docs/index.asciidoc
```diff
@@ -109,14 +128,15 @@ number of cache misses and waste memory.
...
-GeoLite2-City, GeoLite2-Country, GeoLite2-ASN are the free databases from Maxmind that are supported.
-GeoIP2-City, GeoIP2-ISP, GeoIP2-Country are the commercial databases from Maxmind that are supported.
+The path to MaxMind's database file that Logstash should use. The default database is GeoLite2-City.
+GeoLite2-City, GeoLite2-Country, GeoLite2-ASN are the free databases from MaxMind that are supported.
+GeoIP2-City, GeoIP2-ISP, GeoIP2-Country are the commercial databases from MaxMind that are supported.
 
-If not specified, this will default to the GeoLite2 City database that ships
-with Logstash.
+Database auto-update applies to default distribution. When `database` point to user's database path,
+auto-update will be disabled
```
kaisecheng, line 138: Be more explicit here to emphasize no auto-update if `database` is set

docs/index.asciidoc
```diff
@@ -109,14 +128,15 @@ number of cache misses and waste memory.
...
-The path to Maxmind's database file that Logstash should use. The default database is GeoLite2-City.
-GeoLite2-City, GeoLite2-Country, GeoLite2-ASN are the free databases from Maxmind that are supported.
-GeoIP2-City, GeoIP2-ISP, GeoIP2-Country are the commercial databases from Maxmind that are supported.
+The path to MaxMind's database file that Logstash should use. The default database is GeoLite2-City.
+GeoLite2-City, GeoLite2-Country, GeoLite2-ASN are the free databases from MaxMind that are supported.
+GeoIP2-City, GeoIP2-ISP, GeoIP2-Country are the commercial databases from MaxMind that are supported.
 
-If not specified, this will default to the GeoLite2 City database that ships
-with Logstash.
+Database auto-update applies to default distribution. When `database` point to user's database path,
```
karenzone, line 137:
```suggestion
Database auto-update applies to default distribution. When `database` points to user's database path,
```

karenzone (approved):
Left a minor comment inline.  Otherwise, LGTM!

docs/index.asciidoc
```diff
@@ -109,14 +128,15 @@ number of cache misses and waste memory.
...
+auto-update will be disabled
+{logstash-ref}/plugins-filters-geoip.html#plugins-filters-geoip-database_license[see also database license]
```
karenzone, line 139:
```suggestion
See
<<plugins-{type}s-{plugin}-database_license,Database License>> for more information.
```

docs/index.asciidoc
karenzone, line 138:
```suggestion
auto-update will be disabled.
```

Issue #163, opened 2020-02-26
Implement ECS-Compatibility Mode


yaauie:
As a part of the effort to make plugins able to run in an ECS-Compatible manner
by default in an upcoming release of Logstash, this plugin needs to implement
an ECS-Compatibility mode that does not implicitly use fields that conflict
with ECS.

Absent `target` directive, this plugin uses root-level field `geoip`,
which is _undefined_ in ECS and therefore has the possiblity of conflicting
with future versions of ECS. The plugin may benefit from an ECS-Compatibility
mode that requires the target to be set explicitly, ~by using the ECS-compatible
root-level `geo` object~ (edit: _"the geo fields are not expected to be used
directly at the top level."_), or otherwise reduces the risk of future conflict.

Additional analysis will need to be made to ensure the resulting object is also compatible with the [geo field type in ECS](https://www.elastic.co/guide/en/ecs/current/ecs-geo.html).

kares:
Plugin will need to be aligned with [ES output](https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/924) (default template for `geoip: ...` needs to be adjusted)

yaauie:
| DB Field Name    | ECS Equivalent Subfield Name           |
| ---------------- | -------------------------------------- |
| `asn`            | _N/A_                                  |
| `as_org`         | _N/A_                                  |
| `city_name`      | `city_name`                            |
| `country_name`   | `country_name`                         |
| `continent_code` | _N/A_                                  |
| `continent_name` | `continent_name`                       |
| `country_code2`  | `country_iso_code`                     |
| `country_code3`  | _N/A_                                  |
| `ip`             | _N/A_                                  |
| `isp`            | _N/A_                                  |
| `postal_code`    | _N/A_                                  |
| `dma_code`       | _N/A_                                  |
| `region_name`    | `region_name`                          |
| `region_code`    | `region_iso_code`* (`WA` or `US-WA`)?  |
| `timezone`       | _N/A_                                  |
| `location`       | * SEE: `location.lat` & `location.lon` |
| `latitude`       | `location.lat`                         |
| `longitude`      | `location.lon`                         |
| `organization`   | _N/A_                                  |
| _it depends_     | `name`                                 |

jeffrysleddens:
asn and as_org map to the autonomous systems fields (https://www.elastic.co/guide/en/ecs/current/ecs-as.html) as.number and as.organization.name respectivly

PR #179, opened 2021-03-16
add ecs support

kaisecheng:
Elastic Common Schema Support
`target` uses `client` as the default in ECS; legacy keeps `geoip` as default

Co-authored-by: Ry Biesemeyer <ry.biesemeyer@elastic.co>

Fix: https://github.com/logstash-plugins/logstash-filter-geoip/issues/163


yaauie (commented):
First: 🌮🌮🌮 (I meant 🎉  but 🌮  are also good 😂 ). Thank you much for carrying this forward.

I've left a couple notes in-line.

***

I also have thoughts on user experience that I completely failed to define in the original requirements 🤦🏼 :

I believe we should auto-target when the `source` specifies an `ip` sub-field, and should otherwise _require_ a target in ECS mode instead of guessing `client`.

For example, when the user specifies `source => "[host][ip]"` and doesn't provide a `target` we can safely assume `target => "[host]"`, but if they specify a non-ECS `source => "host_ip"` we can't reasonably assume enough in the context of ECS to properly categorize it.

Pseudo-code:
~~~ ruby
def default_ecs_target
  return @source[0...-4] if @source.end_with?('[ip]')

  fail(LogStash::ConfigurationError, "GeoIP Filter in ECS-Compatiblity mode requires a `target` when `source` is not an `ip` sub-field")
end
~~~

spec/filters/geoip_ecs_spec.rb
```diff
@@ -0,0 +1,95 @@
...
+        end
+
+        context "with city database" do
+          let(:options) { common_options }
+
+          it "should return geo in target" do
+            plugin.filter(event)
+
+            expect( event.get ecs_select[disabled: "[#{target}][ip]", v1: "[#{target}][ip]"] ).to eq ip
+            expect( event.get ecs_select[disabled: "[#{target}][country_code3]", v1: "[#{target}][country_code3]"] ).to eq 'US'
```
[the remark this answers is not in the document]
kaisecheng: I remove it in ECS mode, legacy mode shows `country_code3`

lib/logstash/filters/geoip.rb
```diff
@@ -108,10 +113,18 @@ def tag_unsuccessful_lookup(event)
     @tag_on_failure.each{|tag| event.tag(tag)}
   end
 
+  def setup_target_field
+    @target ||= ecs_select[disabled:'geoip', v1:'client']
+
+    if ecs_compatibility != :disabled and !ECS_TARGET_FIELD.include?(@target)
```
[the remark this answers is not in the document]
kaisecheng: good catch!

lib/logstash/filters/geoip.rb
```diff
@@ -108,10 +124,23 @@ def tag_unsuccessful_lookup(event)
...
+      @target ||= 'geoip'
+    else
+      auto_target = @source[0...-4] if @source.end_with?('[ip]') and @source.length > 4
+      @target ||= auto_target || fail(LogStash::ConfigurationError, "GeoIP Filter in ECS-Compatiblity mode "\
+                                            "requires a `target` when `source` is not an `ip` sub-field, eg. [client][ip]")
+
+      normalized_target = (@target.start_with?('[') and @target.end_with?(']'))? @target[1...-1] : @target
+      logger.warn("ECS expect `target` value in #{ECS_TARGET_FIELD}") unless ECS_TARGET_FIELD.include?(normalized_target)
+    end
+  end
```
yaauie, line 138: There is a lot going on with conditional assignments, which could be made a lot simpler by extracting `auto_target_from_source` to its own method.

A couple separate points:

1. We tend to avoid the `and` and `or` keywords in Ruby conditionals (favoring their symbolic counterparts `&&` and `||`) because they have surprising behaviour with regard to precedence (more: https://avdi.codes/how-to-use-rubys-english-andor-operators-without-going-nuts/).
2. I'm wary of just stripping open-and-close brackets to normalize the field reference, because nested fields end up with a funky syntax that would be hard to replicate if/when ECS adds a host-like field that is _not_ at the top-level. Using an anchored-at-both-ends gsub like below will allow us to strip the bracket syntax from top-level fields while leaving nested fields alone.

```suggestion
  def setup_target_field
    if ecs_compatibility == :disabled
      @target ||= 'geoip'
    else
      @target ||= auto_target_from_source!

      # normalize top-level fields to not be bracket-wrapped
      normalized_target = @target.gsub(/\A\[([^\[\]]+)\]\z/,'\1')

      logger.warn("ECS expect `target` value `#{normalized_target}` in #{ECS_TARGET_FIELD}") unless ECS_TARGET_FIELD.include?(normalized_target)
    end
  end

  def auto_target_from_source!
    return @source[0...-4] if @source.end_with?('[ip]') && @source.length > 4
      
    fail(LogStash::ConfigurationError, "GeoIP Filter in ECS-Compatiblity mode "\
                                       "requires a `target` when `source` is not an `ip` sub-field, eg. [client][ip]")
  end
```

spec/filters/geoip_ecs_spec.rb
```diff
@@ -0,0 +1,203 @@
...
+          allow_any_instance_of(described_class).to receive(:ecs_compatibility).and_return(:v1)
+        end
+
+        context "`target` is unset" do
+
+          context "`source` end with [ip]" do
+            let(:event) { LogStash::Event.new("host" => {"ip" => ip}) }
+            let(:options) { common_options.merge({"source" => "[host][ip]"}) }
+
+            it "should use [host] as target" do
```
yaauie, line 125: nit: the specs in documentation format can be made more readable if we don't include implementation-specific stuff in the description.

```suggestion
            it "should use source's parent as target" do
```

spec/filters/geoip_ecs_spec.rb
```diff
@@ -0,0 +1,203 @@
...
+              plugin.filter(event)
+              expect( event.get "[host][geo][country_iso_code]" ).to eq 'US'
+            end
+          end
+
+          context "`source` end with [ip] but `target` does not match ECS template" do
+            let(:event) { LogStash::Event.new("hostname" => {"ip" => ip}) }
+            let(:options) { common_options.merge({"source" => "[hostname][ip]"}) }
+
+            it "should use [hostname] as target with warning" do
```
yaauie, line 136: nit: same as above

```suggestion
            it "should use source's parent as target with warning" do
```

spec/filters/geoip_ecs_spec.rb
```diff
@@ -0,0 +1,203 @@
...
+              expect( event.get "[hostname][geo][country_iso_code]" ).to eq 'US'
+            end
+          end
+
+          context "`source` == [ip]" do
+            let(:event) { LogStash::Event.new("ip" => ip) }
+            let(:options) { common_options.merge({"source" => "[ip]"}) }
+
+            it "should raise error to require `target`" do
+              expect { plugin.register }.to raise_error LogStash::ConfigurationError
```
yaauie, line 149: nit: ensure we aren't failing for some other reason by validating the error message contents with a regexp

```suggestion
              expect { plugin.register }.to raise_error LogStash::ConfigurationError, /requires a `target`/
```

spec/filters/geoip_ecs_spec.rb
```diff
@@ -0,0 +1,203 @@
...
+            end
+          end
+
+          context "`source` not end with [ip]" do
+            let(:event) { LogStash::Event.new("host_ip" => ip) }
+            let(:options) { common_options.merge({"source" => "host_ip"}) }
+
+            it "should raise error to require `target`" do
+              expect { plugin.register }.to raise_error LogStash::ConfigurationError
```
yaauie, line 158: nit: same as above, assert error message contents

```suggestion
              expect { plugin.register }.to raise_error LogStash::ConfigurationError, /requires a `target`/
```

spec/filters/geoip_ecs_spec.rb
```diff
@@ -0,0 +1,203 @@
...
+              plugin.filter(event)
+              expect( event.get "[host][geo][country_iso_code]" ).to eq 'US'
+            end
+          end
+
+          context "`target` does not match ECS template" do
+            let(:target) { 'host_ip' }
+
+            it "should use `target` with warning" do
+              expect(plugin.logger).to receive(:warn)
```
yaauie, line 192: nit: assert the contents of the warning

```suggestion
              expect(plugin.logger).to receive(:warn).with(/ECS expect `target`/)
```

yaauie (commented):
👍🏼 great spec coverage!

I've left a couple nitpicks about constraining assertions in specs, and a suggested refactoring of the target detection to make it easier to read and reason about.

lib/logstash/filters/geoip.rb
kaisecheng: 1 :exploding_head:

docs/index.asciidoc
```diff
@@ -89,6 +89,7 @@ This plugin supports the following configuration options plus the <<plugins-{typ
 [cols="<,<,<",options="header",]
 |=======================================================================
 |Setting |Input type|Required
+| <<plugins-{type}s-{plugin}-ecs_compatibility>> | <<string,string>>|No
```
karenzone, line 92: Doc: Please order alphabetically in both the table and doc description.  Otherwise, build cleanly and LGTM>

karenzone (commented):
Left comment in line.  Otherwise, builds cleanly and LGTM.

yaauie (approved):
:tada: LGTM

kaisecheng:
published at v7.1.0

PR #181, opened 2021-04-29
fix database manager with multiple pipelines and expiry action

kaisecheng:
This PR goes together with another Logstash PR https://github.com/elastic/logstash/pull/12862

This PR does two things
- The geoip DatabaseManager in core is highly coupling to geoip plugin. Since the class change to a singleton instance, this plugin needs to update corresponded part.
- Database expiry action has changed. Instead of stopping the pipeline, the plugin fails itself by adding a tag `_geoip_expired_database` and bypass database lookup.

Fixed: https://github.com/elastic/logstash/issues/12856


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 7.2.0
+  - Support multiple pipelines using the same database [#181](https://github.com/logstash-plugins/logstash-filter-geoip/pull/181)
+  - Add EULA doc
```
kaisecheng, line 3: @jsvd this version only works with Logstash 7.14+. I will modify the description. Do I need to do anything to make sure 7.13.x won't use this version?

jsvd: we can pin this requirement in the gemspec:

```ruby
  # and write a short sentence about why this is pinned, maybe with a link to an issue or comment
  s.add_runtime_dependency "logstash-core", ">= 7.14.0"
```

lib/logstash/filters/geoip.rb
```diff
@@ -163,31 +169,24 @@ def fail_filter
...
+      vendor_path = ::File.expand_path(::File.join("..", "..", "..", "..", "vendor"), __FILE__)
       @database.nil? ? ::File.join(vendor_path, "GeoLite2-#{@default_database_type}.mmdb") : @database
     end
   end
 
   def load_database_manager?
     begin
       require_relative ::File.join(LogStash::Environment::LOGSTASH_HOME, "x-pack", "lib", "filters", "geoip", "database_manager")
-      true
+      LogStash::Filters::Geoip::DatabaseManager.respond_to?(:instance)
```
jsvd, line 189: As an alternative you can check the actual logstash version with

```
irb(main):008:0> LOGSTASH_VERSION
=> "7.13.0"
irb(main):009:0> Gem::Requirement.new(">= 7.14.0").satisfied_by?(Gem::Version.new(LOGSTASH_VERSION))
=> false
```


```ruby
MINIMUM_LOGSTASH_VERSION=">= 7.14.0"

def compatible_logstash_version?
  Gem::Requirement.new(MINIMUM_LOGSTASH_VERSION).satisfied_by?(LOGSTASH_VERSION)
end
```

```suggestion
      compatible_logstash_version?
```

lib/logstash/filters/geoip.rb
```diff
@@ -153,8 +152,15 @@ def setup_filter(database_path)
...
+  def update_filter(action, *args)
+    case
+    when action == :update
+      setup_filter(*args)
+    when action == :expire
+      fail_filter
+    else
+      @logger.warn("invalid callback", :action => action)
+    end
   end
```
jsvd, line 164: minor ruby style suggestion:

```suggestion
    case action
    when :update
      setup_filter(*args)
    when :expire
      fail_filter
    else
      @logger.warn("invalid callback", :action => action)
    end
  end
```

docs/index.asciidoc
```diff
@@ -37,6 +37,25 @@ can be https://dev.maxmind.com/geoip/geoip2/geolite2[downloaded from here].
...
 If you would like to get Autonomous System Number(ASN) information, you can use the GeoLite2-ASN database.
 
+[id="plugins-{type}s-{plugin}-database_license"]
+==== Database License
+
+https://www.maxmind.com[MaxMind] changed from releasing the GeoIP database under
+a Creative Commons (CC) license to a proprietary end-user license agreement
+(EULA).  The MaxMind EULA requires Logstash to update the MaxMind database
+within 30 days of a database update. If Logstash fails to download the database
+for 30 days, it will stop the pipeline in order to maintain compliance.
```
jsvd, line 47:
```suggestion
for 30 days, the geoip filter will stop enriching events in order to maintain compliance. Events will be tagged with `_geoip_expired_database` tag to facilitate the handling of these events.
```

jsvd (approved):
LGTM with a minor change to docs, as we don't yet mention anywhere the tagging of events in the case of an expired database.

docs/index.asciidoc
kaisecheng: There will be a follow-up PR for the doc. I will update this line. thx :)

PR #182, opened 2021-05-10
Change database expire action to fail the pipeline

kaisecheng:
This PR goes together with https://github.com/elastic/logstash/pull/12884
The PR changes the behaviour of database expiry. Instead of stopping the pipeline, it fails the pipeline by adding a tag `_geoip_expired_database` to the event.

Fixed: https://github.com/elastic/logstash/issues/12860

### DIFF
--- a/lib/logstash/filters/geoip.rb
+++ b/lib/logstash/filters/geoip.rb
@@ -112,6 +112,8 @@ class LogStash::Filters::GeoIP < LogStash::Filters::Base
   public
   def filter(event)
     return unless filter?(event)
+    return event.tag("_geoip_expired_database") unless @healthy_database
+
     if @geoipfilter.handleEvent(event)
       filter_matched(event)
     else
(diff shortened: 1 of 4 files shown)


lib/logstash/filters/geoip.rb
```diff
@@ -144,11 +146,22 @@ def auto_target_from_source!
...
     @geoipfilter = org.logstash.filters.geoip.GeoIPFilter.new(@source, @target, @fields, @database, @cache_size, ecs_compatibility.to_s)
   end
 
+  # call by DatabaseManager
+  def expire_action
+    fail_filter
+  end
+
+  def fail_filter
+    @logger.info("geoip plugin stop filtering")
```
jsvd, line 161: I suggest moving this to warn:
```suggestion
    @logger.warn("geoip plugin will stop filtering and will tag all events with the '_geoip_expired_database' tag.")
```

should we check that we are flipping `@health_database` from true to false? or can it happen that every day we're printing the message above ?

kaisecheng: I updated the log message to print the value of health_database explicitly. Nice suggestion

jsvd (approved):
LGTM

Issue #195, opened 2021-12-06
missing ECS fields


emalzer:
**Logstash information**:

1. Logstash version
```
# /usr/share/logstash/bin/logstash --version
Using bundled JDK: /usr/share/logstash/jdk
logstash 7.15.1
```
2. Logstash installation source (e.g. built from source, with a package manager: DEB/RPM, expanded from tar or zip archive, docker)
DEB / Ubuntu 18.04

3. How is Logstash being run (e.g. as a service/service manager: systemd, upstart, etc. Via command line, docker/kubernetes)
systemd

4. How was the Logstash Plugin installed
core plugin

**JVM** (e.g. `java -version`):
```openjdk version "11.0.11" 2021-04-20
OpenJDK Runtime Environment (build 11.0.11+9-Ubuntu-0ubuntu2.18.04)
OpenJDK 64-Bit Server VM (build 11.0.11+9-Ubuntu-0ubuntu2.18.04, mixed mode, sharing)
```

**OS version** (`uname -a` if on a Unix-like system):
```
Linux minion-em-elk01 4.15.0-162-generic #170-Ubuntu SMP Mon Oct 18 11:38:05 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
```

**Description of the problem including expected versus actual behaviour**:
The current ECS scheme (1.12) uses the field geo.region_iso_code field, the geoip filter plugin actually uses the geo.region_code field, which does not comply with the ECS scheme. Also it inserts the country_code2 and country_code3 (which is fine for me :) ), but not the geo.country_iso_code.

**Steps to reproduce**:

```
root@minion-em-elk01:/usr/share/logstash# ./bin/logstash -e 'input { stdin { codec => json } } filter { geoip { source => "ip" target => "client" } } output { stdout {} }'
....
{ "ip": "52.168.174.229" }
{
            "ip" => "52.168.174.229",
      "@version" => "1",
    "@timestamp" => 2021-12-06T09:32:26.081Z,
          "host" => "minion-em-elk01",
        "client" => {
              "latitude" => 37.9254,
              "location" => {
            "lat" => 37.9254,
            "lon" => -76.8591
        },
          "country_name" => "United States",
              "dma_code" => 556,
             "longitude" => -76.8591,
           "postal_code" => "22560",
         "country_code3" => "US",
         "country_code2" => "US",
           "region_code" => "VA",
        "continent_code" => "NA",
             "city_name" => "Tappahannock",
[...]

PR #205, opened 2022-01-13
Fix ECS field region_iso_code

kaisecheng:
This commit changes the ECS field from `geo.region_code` to `geo.region_iso_code`

Before this change, `52.168.174.229` gives `region_code`: `VA`

The source of `region_code` is Maxmind `iso_code`. `region_code` is not a ECS field but `region_iso_code` is.
In ECS example the value of`region_iso_code` is `CA-QC`, which seems to be a combination of country iso code and region iso code. So, the current `region_code` is not the same as ECS schema `region_iso_code`.

Questions: 
1. Can we change `geo.region_code` to `geo.region_iso_code` or keep it as is?
2. Should we use country iso code and region iso code to build ECS `regsion_iso_code`?

Fixed: #195


kaisecheng:
need to combine country iso code and region iso code instead of renaming the field

PR #206, opened 2022-01-25
ECS: support composite region_iso_code

yaauie:
Adds support for ECS's composite `region_iso_code`, which combines the ISO 2-char country code with the ISO 2-char region code.

When running in ECS mode with these changes, we _prefer_ the `region_iso_code` to the non-ECS `region_code`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 7.2.11
+ - Improved compatibility with the Elastic Common Schema [#206](https://github.com/logstash-plugins/logstash-filter-geoip/pull/206)
+   - Added support for ECS's composite `region_iso_code`
+   - [DOC] Improve ECS-related documentation
+
 ## 7.2.10
   - [DOC] Air-gapped environment requires both ASN and City databases [#204](https://github.com/logstash-plugins/logstash-filter-geoip/pull/204)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## 7.2.11
  - Improved compatibility with the Elastic Common Schema [#206](https://github.com/logstash-plugins/logstash-filter-geoip/pull/206)
-   - Added support for ECS's composite `region_iso_code`
+   - Added support for ECS's composite `region_iso_code` (`US-WA`), which _replaces_ the non-ECS `region_code` (`WA`) as a default field with City databases. To get the stand-alone `region_code` in ECS mode, you must include it in the `fields` directive. 
    - [DOC] Improve ECS-related documentation
 
 ## 7.2.10
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -172,7 +172,8 @@ Example response:
 [id="plugins-{type}s-{plugin}-field-mapping"]
 ==== Field mapping
 
-When this plugin is run _without_ <<plugins-{type}s-{plugin}-ecs_compatibility>>, the MaxMind DB's fields are added directly to the <<plugins-{type}s-{plugin}-target>>, but when ECS compatibility is enabled, the fields are structured to fit into an ECS shape.
+When this plugin is run with <<plugins-{type}s-{plugin}-ecs_compatibility>> disabled, the MaxMind DB's fields are added directly to the <<plugins-{type}s-{plugin}-target>>.
+When ECS compatibility is enabled, the fields are structured to fit into an ECS shape.
 
 [cols="3,5,3"]
 |===========================
@@ -217,8 +218,8 @@ NOTE: `*` indicates a composite field, which is only populated if GeoIP lookup r
 
 When using a City database, the enrichment is aborted if no latitude/longitude pair is available.
 
-The `location` field combines the latitude and longitude into a structure called http://geojson.org/geojson-spec.html[GeoJSON].
-When using a default <<plugins-{type}s-{plugin}-target>>, the templates provided by the {logstash-ref}/plugins-outputs-elasticsearch.html[elasticsearch output] maps the field to an {ref}/geo-point.html[Elasticsearch Geo_point datatype].
+The `location` field combines the latitude and longitude into a structure called https://datatracker.ietf.org/doc/html/rfc7946[GeoJSON].
+When you are using a default <<plugins-{type}s-{plugin}-target>>, the templates provided by the {logstash-ref}/plugins-outputs-elasticsearch.html[elasticsearch output] map the field to an {ref}/geo-point.html[Elasticsearch Geo_point datatype].
 
 As this field is a `geo_point` _and_ it is still valid GeoJSON, you get
 the awesomeness of Elasticsearch's geospatial query, facet and filter functions
@@ -291,8 +292,8 @@ The default database is `GeoLite2-City`.
 This plugin supports several free databases (`GeoLite2-City`, `GeoLite2-Country`, `GeoLite2-ASN`)
 and a selection of commercially-licensed databases (`GeoIP2-City`, `GeoIP2-ISP`, `GeoIP2-Country`).
 
-Database auto-update applies to default distribution.
-When `database` points to user's database path, auto-update will be disabled.
+Database auto-update applies to the default distribution.
+When `database` points to user's database path, auto-update is disabled.
 See <<plugins-{type}s-{plugin}-database_license,Database License>> for more information.
 
 [id="plugins-{type}s-{plugin}-default_database_type"]
@@ -323,7 +324,7 @@ For a complete list of available fields and how they map to an event's structure
 * Value type is <<string,string>>
 * Supported values are:
 ** `disabled`: unstructured geo data added at root level
-** `v1`, `v8`: uses fields that are compatible with Elastic Common Schema (for example, `[client][geo][country_name]`; see <<plugins-{type}s-{plugin}-field-mapping,field mapping>>)
+** `v1`, `v8`: use fields that are compatible with Elastic Common Schema. Example: `[client][geo][country_name]`. See <<plugins-{type}s-{plugin}-field-mapping,field mapping>> for more info. 
 * Default value depends on which version of Logstash is running:
 ** When Logstash provides a `pipeline.ecs_compatibility` setting, its value is used as the default
 ** Otherwise, the default value is `disabled`.
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -169,14 +169,55 @@ Example response:
 }
 --------------------------------------------------
 
+[id="plugins-{type}s-{plugin}-field-mapping"]
+==== Field mapping
+
+When this plugin is run _without_ <<plugins-{type}s-{plugin}-ecs_compatibility>>, the MaxMind DB's fields are added directly to the <<plugins-{type}s-{plugin}-target>>, but when ECS compatibility is enabled, the fields are structured to fit into an ECS shape.
+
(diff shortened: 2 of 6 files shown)


src/main/java/org/logstash/filters/geoip/Fields.java
```diff
@@ -96,6 +97,14 @@ public String getFieldReferenceECSv1() {
       Fields.COUNTRY_CODE3, Fields.IP, Fields.POSTAL_CODE, Fields.DMA_CODE, Fields.REGION_NAME,
       Fields.REGION_CODE, Fields.TIMEZONE, Fields.LOCATION, Fields.LATITUDE, Fields.LONGITUDE);
 
+  // When ECS is enabled, the composite REGION_ISO_CODE field is preferred to separate REGION_CODE
+  static final EnumSet<Fields> DEFAULT_ECS_CITY_FIELDS;
+  static {
+    DEFAULT_ECS_CITY_FIELDS = EnumSet.copyOf(DEFAULT_CITY_FIELDS);
+    DEFAULT_ECS_CITY_FIELDS.remove(REGION_CODE);
```
yaauie, line 104: This is perhaps a breaking change from a certain perspective, for a user who was relying on the non-ECS `[geo][region_code]` being populated in ECS mode. I'm inclined to call this a bugfix, since users have to explicitly opt into ECS to hit this behaviour.

But if we want to continue populating `[geo][region_code]` in ECS mode, this change and a small change to the specs would suffice:

```suggestion
```

spec/filters/geoip_ecs_spec.rb
```diff
@@ -36,15 +40,23 @@
...
             if ecs_select.active_mode == :disabled
               expect( event.get "[#{target}][country_code3]" ).to eq 'US'
+              expect( event.get "[#{target}][region_code]" ).to eq 'MA'
+              expect( event.get "[#{target}][region_iso_code]" ).to be_nil
             else
               expect( event.get "[#{target}][geo][country_code3]" ).to be_nil
               expect( event.get "[#{target}][country_code3]" ).to be_nil
+              expect( event.get "[#{target}][geo][region_iso_code]" ).to eq 'US-MA'
+              expect( event.get "[#{target}][region_code]" ).to be_nil
             end
```
yaauie, line 58: If we _want_ to keep populating the non-ECS `[geo][region_code]`, this change is also needed:

```suggestion
            if ecs_select.active_mode == :disabled
              expect( event.get "[#{target}][country_code3]" ).to eq 'US'
              expect( event.get "[#{target}][region_code]" ).to eq 'MA'
              expect( event.get "[#{target}][region_iso_code]" ).to be_nil
            else
              expect( event.get "[#{target}][geo][country_code3]" ).to be_nil
              expect( event.get "[#{target}][country_code3]" ).to be_nil
              expect( event.get "[#{target}][geo][region_iso_code]" ).to eq 'US-MA'
              expect( event.get "[#{target}][geo][region_code]" ).to eq 'MA'
            end
```

docs/index.asciidoc
```diff
@@ -169,14 +169,56 @@ Example response:
...
+
+| `domain`           | `[domain]`                 | `example.com`
+
+| `asn`              | `[as][number]`             | `98765`
+| `as_org`           | `[as][organization][name]` | `Elastic, NV`
+
+| `isp`              | `[mmdb][isp]`              | `InterLink Supra LLC`
+| `dma_code`         | `[mmdb][dma_code]`         | `819`
+| `organization`     | `[mmdb][organization]`     | `Elastic, NV`
+|===========================
```
karenzone, line 212: Love the table. Nice presentation and it renders well, too.

docs/index.asciidoc
```diff
@@ -169,14 +169,56 @@ Example response:
...
 
-A `[geoip][location]` field is created if
-the GeoIP lookup returns a latitude and longitude. The field is stored in
-http://geojson.org/geojson-spec.html[GeoJSON] format. Additionally,
-the default Elasticsearch template provided with the
-{logstash-ref}/plugins-outputs-elasticsearch.html[elasticsearch output] maps
-the `[geoip][location]` field to an {ref}/geo-point.html[Elasticsearch Geo_point datatype].
+When using a City database, the enrichment is aborted if no latitude/longitude pair is available.
+
+The `location` field combines the latitude and longitude into a structure called http://geojson.org/geojson-spec.html[GeoJSON].
```
karenzone, line 220: This link (http://geojson.org/geojson-spec.html) resolves, but it says the document is obsolete. It recommends https://datatracker.ietf.org/doc/html/rfc7946 instead. Should we update or point to something else?

docs/index.asciidoc
```diff
@@ -169,14 +169,56 @@ Example response:
...
+When using a default <<plugins-{type}s-{plugin}-target>>, the templates provided by the {logstash-ref}/plugins-outputs-elasticsearch.html[elasticsearch output] maps the field to an {ref}/geo-point.html[Elasticsearch Geo_point datatype].
```
karenzone, line 221:
```suggestion
When you are using a default <<plugins-{type}s-{plugin}-target>>, the templates provided by the {logstash-ref}/plugins-outputs-elasticsearch.html[elasticsearch output] map the field to an {ref}/geo-point.html[Elasticsearch Geo_point datatype].
```

karenzone: In this construction, "templates" becomes the subject, which doesn't really make sense. Suggesting "you..." or perhaps "the output..." 
The verb "map" should agree with "templates" (instead of elasticsearch output).

docs/index.asciidoc
```diff
@@ -242,16 +284,16 @@ number of cache misses and waste memory.
...
+The path to MaxMind's database file that Logstash should use.
+The default database is `GeoLite2-City`.
+This plugin supports several free databases (`GeoLite2-City`, `GeoLite2-Country`, `GeoLite2-ASN`)
+and a selection of commercially-licensed databases (`GeoIP2-City`, `GeoIP2-ISP`, `GeoIP2-Country`).
 
-Database auto-update applies to default distribution. When `database` points to user's database path,
-auto-update will be disabled.
-See
-<<plugins-{type}s-{plugin}-database_license,Database License>> for more information.
+Database auto-update applies to default distribution.
```
karenzone, line 294:
```suggestion
Database auto-update applies to the default distribution.
```

docs/index.asciidoc
```diff
@@ -242,16 +284,16 @@ number of cache misses and waste memory.
...
+When `database` points to user's database path, auto-update will be disabled.
```
karenzone, line 295:
```suggestion
When `database` points to user's database path, auto-update is disabled.
```

docs/index.asciidoc
```diff
@@ -270,21 +312,18 @@ This plugin now includes both the GeoLite2-City and GeoLite2-ASN databases.  If
...
+For a complete list of available fields and how they map to an event's structure, see <<plugins-{type}s-{plugin}-field-mapping,field mapping>>.
 
 [id="plugins-{type}s-{plugin}-ecs_compatibility"]
 ===== `ecs_compatibility`
 
 * Value type is <<string,string>>
 * Supported values are:
 ** `disabled`: unstructured geo data added at root level
-** `v1`, `v8`: uses fields that are compatible with Elastic Common Schema (for example, `[client][geo][country_name]`)
+** `v1`, `v8`: uses fields that are compatible with Elastic Common Schema (for example, `[client][geo][country_name]`; see <<plugins-{type}s-{plugin}-field-mapping,field mapping>>)
```
karenzone, line 326:
```suggestion
** `v1`, `v8`: use fields that are compatible with Elastic Common Schema. Example: `[client][geo][country_name]`. See <<plugins-{type}s-{plugin}-field-mapping,field mapping>> for more info. 
```

karenzone: Minor tweak for easier reading and translation

karenzone (commented):
Nice improvements. Builds cleanly and the table renders as it should.  I left some suggestion inline related to readability, translation, tense, subject/verb agreement, etc. Please let me know if you would like to discuss. 

Sounds like you're on a tight timeline. If you want to proceed with what you have, you have my LGTM.  I can do a follow up PR with these tweaks.

docs/index.asciidoc
yaauie, line 220:
```suggestion
The `location` field combines the latitude and longitude into a structure called https://datatracker.ietf.org/doc/html/rfc7946[GeoJSON].
```

yaauie: IETF standards track is better than a private `.org` in my book 👍🏼

karenzone (approved):
LGTM

src/main/java/org/logstash/filters/geoip/Fields.java
kaisecheng: With ECS default on in 8.x, this is a breaking change. Users do not have the same data of `region_code` or replacement. Thinking to have both `region_code` and `region_iso_code`, but in some sense, the data of `region_code` is also a kind of iso code. Having both seems to be confusing. +1 to the breaking change.

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 7.2.11
+ - Improved compatibility with the Elastic Common Schema [#206](https://github.com/logstash-plugins/logstash-filter-geoip/pull/206)
+   - Added support for ECS's composite `region_iso_code`
```
kaisecheng, line 3: let reader know `region_code` is removed in ECS

kaisecheng (commented):
Thanks for adding the missing ECS doc 💮 I support the change of removing `[geo][region_code]` in ECS mode. Left two comments on the code side

src/main/java/org/logstash/filters/geoip/GeoIPFilter.java
```diff
@@ -311,6 +311,12 @@ private boolean applyGeoData(Map<Fields, Object> geoData, Event event) {
             geoData.put(Fields.REGION_CODE, subdivisionCode);
           }
           break;
+        case REGION_ISO_CODE:
+          String countryCodeForRegion = country.getIsoCode();
+          String regionCode2 = subdivision.getIsoCode();
+          if (countryCodeForRegion != null && regionCode2 != null) {
+            geoData.put(Fields.REGION_ISO_CODE, String.format("%s-%s", countryCodeForRegion, regionCode2));
+          }
```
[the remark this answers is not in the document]
yaauie:
```suggestion
          }
          break;
```

CHANGELOG.md
yaauie:
```suggestion
   - Added support for ECS's composite `region_iso_code` (`US-WA`), which _replaces_ the non-ECS `region_code` (`WA`). 
```

src/main/java/org/logstash/filters/geoip/Fields.java
yaauie: > With ECS default on in 8.x, this is a breaking change

A user upgrading from Logstash 7 to Logstash 8 should be expecting to be consuming breaking changes, and the path to ECS provides appropriate guidance on how to keep a plugin in non-ECS mode (hint: set `pipeline.ecs_compatibility: disabled` for the pipeline to lock in 7.x behaviour in _all_ its plugins).

It is _a little_ breaking for users who _already_ opt into ECS for the pipeline or an instance of this plugin, but it is also a bugfix since users who are already going out of their way to get ECS have a reasonable expectation that the result will be ECS (which `[geo][region_code]` is not).

PR #208, opened 2022-04-01
Added support for DB-IP Location+ISP (Enterprise) databases, [#174]

pepl:
Tested to be working with the commercial DB-IP Location-ISP database.

I do not have a that version of database available, but also the MaxMind Enterprise MMDB should work accordingly if that specific metadata type string would be added to the list of known types.


pepl:
Any news on this?

PR #210, opened 2022-12-27
[DOC] Add documentation for database downloader setting

edmocosta:
Related issue: [#14823](https://github.com/elastic/logstash/pull/14823)

### DIFF
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -47,10 +47,9 @@ within 30 days of a database update.
 
 The GeoIP filter plugin can manage the database for users running the Logstash default
 distribution, or you can manage
-database updates on your own. The behavior is controlled by the `database` setting.
-When you use the default `database` setting, the auto-update feature ensures that the plugin is
+the updates on your own, becoming also responsible for maintaining compliance. The behavior is controlled by the `database` setting.
+When you use the default `database` setting, and the auto-update feature is enabled, Logstash ensures that the plugin is
 using the latest version of the database.
-Otherwise, you are responsible for maintaining compliance.
 
 The Logstash open source distribution uses the MaxMind Creative Commons license
 database by default.
@@ -59,11 +58,8 @@ database by default.
 ==== Database Auto-update
 
 This plugin bundles Creative Commons (CC) license databases.
-Logstash checks for database updates every day. It downloads the latest and can replace the old database
+By default, Logstash checks for database updates every day and upgrades from the CC database licence to the EULA version automatically. It downloads the latest version and replaces old databases
 while the plugin is running.
-After Logstash downloads EULA license databases, it will not fallback to CC license databases.
-
-NOTE: If the database has never been updated successfully, as in air-gapped environments, Logstash can use CC license databases indefinitely.
 
 After Logstash has switched to a EULA licensed database, the geoip filter will
 stop enriching events in order to maintain compliance if Logstash fails to
@@ -74,10 +70,9 @@ TIP: When possible, allow Logstash to access the internet to download databases 
 
 **Disable the auto-update feature**
 
-If you work in air-gapped environment, and want to disable the auto-update feature from Logstash, you can do so by setting the `xpack.geoip.db.auto_update`
-configuration to `false` in `logstash.yml`.
+If you work in air-gapped environment and want to disable the database auto-update feature, set the `xpack.geoip.downloader.enabled` value to `false` in `logstash.yml`.
 
-When disabled, a valid `database` location is required, and you are responsible for <<plugins-{type}s-{plugin}-manage_update,manage your own database updates>> and <<plugins-{type}s-{plugin}-database_license,maintaining compliance>>.
+When disabled, Logstash uses the Creative Commons (CC) license databases indefinitely, and any previously downloaded version of the EULA databases will be deleted.
 
 [id="plugins-{type}s-{plugin}-manage_update"]
 ==== Manage your own database updates
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -47,9 +47,10 @@ within 30 days of a database update.
 
 The GeoIP filter plugin can manage the database for users running the Logstash default
 distribution, or you can manage
-the updates on your own, becoming also responsible for maintaining compliance. The behavior is controlled by the `database` setting.
-When you use the default `database` setting, and the auto-update feature is enabled, Logstash ensures that the plugin is
+database updates on your own. The behavior is controlled by the `database` setting and by the auto-update feature.
+When you use the default `database` setting and the auto-update feature is enabled, Logstash ensures that the plugin is
 using the latest version of the database.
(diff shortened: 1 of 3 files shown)


docs/index.asciidoc
```diff
@@ -72,6 +72,13 @@ Events will be tagged with `_geoip_expired_database` tag to facilitate the handl
 
 TIP: When possible, allow Logstash to access the internet to download databases so that they are always up-to-date.
 
+**Disable the auto-update feature**
+
+If you work in air-gapped environment, and want to disable the auto-update feature from Logstash, you can do so by setting the `xpack.geoip.db.auto_update`
+configuration to `false` in `logstash.yml`.
```
karenzone, line 78:
```suggestion
If you work in air-gapped environment and want to disable the database auto-update feature, set the `xpack.geoip.db.auto_update` value to `false` in `logstash.yml`.
```

docs/index.asciidoc
```diff
@@ -72,6 +72,13 @@ Events will be tagged with `_geoip_expired_database` tag to facilitate the handl
...
+
+When disabled, a valid `database` location is required, and you are responsible for <<plugins-{type}s-{plugin}-manage_update,manage your own database updates>> and <<plugins-{type}s-{plugin}-database_license,maintaining compliance>>.
```
karenzone, line 80:
```suggestion
When automatic updates are disabled, a valid `database` location is required.
You are responsible for managing your own database updates and <<plugins-{type}s-{plugin}-database_license,maintaining license compliance>>.
```

karenzone: Consider removing the link to "maintaining compliance."  It's the very next topic and clicking the link doesn't really take you anywhere.

karenzone (commented):
@edmocosta, I left some comments for your consideration. You can decide. :-) 
Thanks for adding this helpful info!

edmocosta:
Hi @karenzone! Thank you very much for reviewing! 
I've just updated the docs to reflect recent changes discussed on the [#14823](https://github.com/elastic/logstash/pull/14823) PR. Could you please take a look at it again?

docs/index.asciidoc
```diff
@@ -47,10 +47,9 @@ within 30 days of a database update.
 
 The GeoIP filter plugin can manage the database for users running the Logstash default
 distribution, or you can manage
-database updates on your own. The behavior is controlled by the `database` setting.
-When you use the default `database` setting, the auto-update feature ensures that the plugin is
+the updates on your own, becoming also responsible for maintaining compliance. The behavior is controlled by the `database` setting.
```
karenzone, line 50: I'd like to see the note about responsibility moved back to the end of the paragraph.  People tend to read headings, skim through the middle, and read the last sentence. I'm concerned that this important information about responsibility shifting to the user will be overlooked in its current location.

docs/index.asciidoc
```diff
@@ -47,10 +47,9 @@ within 30 days of a database update.
...
+When you use the default `database` setting, and the auto-update feature is enabled, Logstash ensures that the plugin is
```
karenzone, line 51:
```suggestion
When you use the default `database` setting and the auto-update feature is enabled, Logstash ensures that the plugin is
```

karenzone: Nice clarification! Thanks for adding this detail.

docs/index.asciidoc
```diff
@@ -59,11 +58,8 @@ database by default.
 ==== Database Auto-update
 
 This plugin bundles Creative Commons (CC) license databases.
-Logstash checks for database updates every day. It downloads the latest and can replace the old database
+By default, Logstash checks for database updates every day and upgrades from the CC database licence to the EULA version automatically. It downloads the latest version and replaces old databases
```
karenzone, line 61: Has the behavior changed from the previous version of the docs?  

This wording is confusing because it's mixing a daily update with a one time conversion to the EULA license.

karenzone (commented):
Left some comments for consideration. Let me know if you'd like to discuss.

docs/index.asciidoc
edmocosta: Hum, that's true, this behavior didn't actually change from the previous version. Considering we have now an option to disable the auto-update, I tried to explain here that the license also auto-upgrades from CC to EULA when the download is successfully completed. 

What about changing it to something like: 
```
When the auto-update feature is enabled, Logstash checks for database updates every day. It downloads the latest and can replace the old database
```

In that case, I could also restore the removed NOTE to something like:
```
NOTE: If the auto-update feature is disabled or the database has never been updated successfully, as in air-gapped environments, Logstash can use CC license databases indefinitely.
```

What do you think?

docs/index.asciidoc
edmocosta: Thank you for the clarification, that's nice to know! I can maybe reword line 50 a bit to include the auto-update and move everything back to the end of the paragraph as you suggested.

```
database updates on your own. The behavior is controlled by the `database` setting and by the auto-update feature.
```

karenzone: Sounds good.

docs/index.asciidoc
karenzone: Please remember to either commit the suggestion above (removing first comma) or make the change on your own.

docs/index.asciidoc
karenzone: I like where you're heading with this.  If I'm reading your proposed changes correctly, we still need to account for the license conversion.  Something like: 

`If the auto updates feature is enabled, Logstash upgrades from the CC database license to the EULA version on the first download. `

docs/index.asciidoc
```diff
@@ -59,19 +59,26 @@ database by default.
 ==== Database Auto-update
 
 This plugin bundles Creative Commons (CC) license databases.
-Logstash checks for database updates every day. It downloads the latest and can replace the old database
+If the auto-update feature is enabled, Logstash checks for database updates every day. It downloads the latest and can replace the old database
```
karenzone, line 62:
```suggestion
If the auto-update feature is enabled in `logstash.yml`(as it is by default), Logstash checks for database updates every day. It downloads the latest and can replace the old database
```

karenzone: This is a somewhat unusual situation in that we're talking about related settings that are in two different places (logstash.yml and plugin settings).  It would be good to indicate that the auto-update setting is something else and is enabled by default early on.  The rest of the section makes more sense after we set that initial context.

Otherwise, unless somebody reads the "Disable the auto-update feature" setting info, they won't have this piece of information.

karenzone (commented):
My github session expired mid-review and blew away some comments/explanations. I hope that no context is lost. 

I left some notes about being more clear that the `auto-update` and `database` settings are actually set in two different places--the plugin and logstash.yml.

karenzone (approved):
Builds cleanly, renders as expected, and LGTM! Thanks for adding this feature and clarifying doc.

Issue #212, opened 2023-01-30
Enable the use of MaxMind GeoIP2-Anonymous-IP database in the plugin


RiverCitySecDevOps:
An Anonymous IP database PR from 2018 was created: https://github.com/logstash-plugins/logstash-filter-geoip/pull/141

This was never merged back into the plugin. Support for the Anonymous IP database would be greatly appreciated and help support the paid MaxMind databases .

Usage of the database can be found [in MaxMind's GitHub](https://github.com/maxmind/GeoIP2-java#anonymous-ip) and full documentation can be found on [MaxMind's GeoIP2 Anonymous IP Databases page](https://dev.maxmind.com/geoip/docs/databases/anonymous-ip?lang=en#binary-database)

benatbermejo:
+1 Would be nice for fraud detection use case.

Issue #153, opened 2019-05-08
Add support for GeoIP2-Enterprise database


kvmuralidhar:
As written in comment to geoip.rb "The path to the GeoLite2 database file which Logstash should use. City and ASN databases are supported.".

Is there a plan to support GeoIP2-Enterprise database?

Thanks
Murali

kvmuralidhar:
Is there a plan to support GeoIP2-Enterprise database?

bcbee:
Hey @kvmuralidhar!

This plugin already supports MaxMind GeoIP2-Enterprise City and ISP databases. Here is a sample configuration:

```
geoip {
    source => "requestIPAddress"
    database => "/usr/share/logstash/geoip2/GeoIP2-City.mmdb"
    fields => [ "city_name", "country_code2", "region_name", "location" ]
}
geoip {
    source => "requestIPAddress"
    database => "/usr/share/logstash/geoip2/GeoIP2-ISP.mmdb"
    default_database_type => "ASN"
    fields => [ "isp", "organization", "autonomous_system_organization", "autonomous_system_number" ]
}
```

**Docs reference:** https://www.elastic.co/guide/en/logstash/current/plugins-filters-geoip.html#_supported_databases
**Full list of fields:** https://github.com/logstash-plugins/logstash-filter-geoip/blob/master/src/main/java/org/logstash/filters/Fields.java

kvmuralidhar:
@bcbee,

Thank you for your update.  I was looking for support to GeoIP2-Enterprise db since it has additional information for a given ip address.

```
'connection_type': 'Corporate', 
'domain': 'googlebot.com', 
'user_type': 'search_engine_spider'
```

PR #223, opened 2024-05-15
Add support for MaxMind GeoIP2 Enterprise and Anonymous-IP databases

edmocosta:
- Added support for MaxMind GeoIP2 Enterprise and Anonymous-IP databases.
- Changed the database type detection to use the same strategy as [com.maxmind.geoip2.DatabaseReader#getDatabaseType](https://github.com/maxmind/GeoIP2-java/blob/main/src/main/java/com/maxmind/geoip2/DatabaseReader.java#L116), so it doesn't rely on specific database types.
- Updated MaxMind dependencies.
- Added unit tests for the Java classes, including the verification of the plugin's basic operations against all supported databases.
- The gem version is now read from the `VERSION` file.

---
Closes: https://github.com/logstash-plugins/logstash-filter-geoip/issues/153
Closes: https://github.com/logstash-plugins/logstash-filter-geoip/issues/212


src/main/java/org/logstash/filters/geoip/Databases.java
```diff
@@ -0,0 +1,126 @@
...
+            return Databases.ASN;
+        } else if (type.contains(ISP.databaseType)) {
+            return Databases.ISP;
+        } else if (type.contains(ENTERPRISE.databaseType)) {
+            return Databases.ENTERPRISE;
+        } else if (type.contains(ANONYMOUS_IP.databaseType)) {
+            return Databases.ANONYMOUS_IP;
+        }
+
+        // Maybe, we should throw and exception here instead of use an unknown database type.
```
edmocosta, line 120: Any thoughts on this?

src/main/java/org/logstash/filters/geoip/Fields.java
```diff
@@ -44,7 +45,14 @@ enum Fields {
   LOCATION("geo.location", "location"),
   LATITUDE("geo.location.lat", "latitude"),
   LONGITUDE("geo.location.lon", "longitude"),
-  ORGANIZATION("mmdb.organization", "organization");
+  ORGANIZATION("mmdb.organization", "organization"),
+  NETWORK("traits.network", "network"),
```
edmocosta, line 49: Couldn't find any matching names on the ECS schema. I've ended up using the `treats` prefix, as those fields are related to the IP's address characteristics, and are not about the connection, network, etc.
Another option that crossed my mind was using the `is_` prefix, for example `is_anonymous_vpn` , which I didn't implement to keep it similar to the ES geoip field names. 

Do you have any suggestions on this?

src/main/java/org/logstash/filters/geoip/Databases.java
kaisecheng: If the pipeline have multiple geoip-filters and one of it is UNKNOWN, throwing exception in register phase will cause pipeline restart. I prefer to keep backward compatible and add a log to say the current database type and give warning if it is UNKNOWN

src/main/java/org/logstash/filters/geoip/Fields.java
kaisecheng: Looking at the ECS guideline for [custom fields](https://www.elastic.co/guide/en/ecs/current/ecs-custom-fields-in-ecs.html#_proper_names) and the purpose of Anonymous IP database, the name comes across me is `anonymous`.

`anonymous.network`, `anonymous.hosting_provider` ...

kaisecheng (commented):
woohoo! A lot good stuffs in this PR. Thanks for adding test databases 🎉 , fixing the missing close() and updating many legacy parts in this plugin.

My comments are all minor about log and naming. The two new databases run as expected

src/main/java/org/logstash/filters/geoip/Databases.java
```diff
@@ -0,0 +1,126 @@
+package org.logstash.filters.geoip;
+
+import java.util.Collections;
+import java.util.EnumSet;
+import java.util.Set;
+
+enum Databases {
```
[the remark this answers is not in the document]
edmocosta: I've used plural here do keep the existing `Fields` pattern, but I'd definitely be glad changing them to singular! :)

src/main/java/org/logstash/filters/geoip/Fields.java
edmocosta: Yes, I've initially named them using the `anonymous`/`anom` prefix, but TBH I wasn't convinced considering those fields are actually IP's characteristics/traits, for example:
The `network` field is the IP's CIDR ( e.g:  `74.209.16.0/20`), and is more about the network than anonymity.
The other fields are basically traits of the requested IP address, and indicates if the IP **belongs**/**is** a VPN, Tor, public proxy, or hosting provider, etc. Finally, I ended up using the same name as MaxMind does ([traits](https://github.com/maxmind/GeoIP2-java/blob/main/src/main/java/com/maxmind/geoip2/record/Traits.java)), so we could group these and future IP's characteristics fields under this prefix. WDYT?

src/main/java/org/logstash/filters/geoip/Databases.java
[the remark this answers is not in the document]
edmocosta: Done ✅

src/main/java/org/logstash/filters/geoip/Databases.java
edmocosta: Done ✅

src/main/java/org/logstash/filters/geoip/Fields.java
kaisecheng: right, `anonymous` doesn't sound a fit. My main concern is neither `anonymous` nor `traits` hint me this is about IP/ network. How about `ip_traits`?

edmocosta: Done! could you please have a second look (https://github.com/logstash-plugins/logstash-filter-geoip/pull/223/commits/75a85bc1634f28335eec4cbfe4c878a32fa52ba7)? In addition to that, I've downgraded back to 2.x the MaxMind libraries (https://github.com/logstash-plugins/logstash-filter-geoip/pull/223/commits/5b1a40009a68d7b12cdc2f8d76eca73f8c8ad8ad), so we don't need to enforce Java 11, avoiding compatibility issues with LS 7 & Java 8.

src/main/java/org/logstash/filters/geoip/Databases.java
kaisecheng: Have you added a log to user to warn them the database type is unknown?

kaisecheng: Done 👍

kaisecheng (approved):
🚀 LGTM. good catch on supporting jvm 8.

src/main/java/org/logstash/filters/geoip/Fields.java
mcauto: @edmocosta 

It seems that there is no code using `Field.NETWORK`. How should I configure it to use `ip_traits.network`?

PR #224, opened 2024-05-15
Upgrade Gradle to 8.7 and fix deprecations

andsel:
Update Gradle to 8.7 and fixes deprecation warnings.

In particular this commit:
- avoid to set `project.archivesBaseName` property, and use the new syntax
- avoid to read `project.archivesBaseName` property and leverage a local variable to store the artifact name


jsvd (approved):
LGTM

PR #228, opened 2024-11-14
Update .travis.yml to ajust to new release cadence

jsvd:


donoghuc:
Hmm looks like for `8.future` we are missing this transformation:
```
#1 [logstash internal] load build definition from Dockerfile
#1 transferring dockerfile: 1.94kB done
#1 WARN: InvalidDefaultArgInFrom: Default value for ARG docker.elastic.co/logstash/logstash${DISTRIBUTION_SUFFIX}:${ELASTIC_STACK_VERSION} results in empty or invalid base image name (line 3)
#1 DONE 0.0s
#2 [logstash internal] load metadata for docker.elastic.co/logstash/logstash:8.future
#2 ERROR: docker.elastic.co/logstash/logstash:8.future: not found
```
https://github.com/logstash-plugins/.ci/blob/670ab81da4e9e6878fafaff72cf79df3bb8498e3/docker-setup.sh#L47-L57

jsvd:
allowing a 8.future stack version to not exist is not ideal but acceptable, so it should not block the build from being green: https://github.com/logstash-plugins/.ci/pull/68/files

donoghuc:
Agree. Looks like that PR is merged. If you mark this as "ready for review" instead of "draft" will that trigger a fresh travis run? If we go green on that we should be g2g

jsvd:
I manually requested https://app.travis-ci.com/github/logstash-plugins/logstash-filter-geoip/builds/273267562, let's see..

jsvd:
@donoghuc I've simplified the travis.yml by only making it add the extra oss tests, leaving the default distribution tests to be read by the shared .ci definition. everything's green now.

donoghuc (approved):
+1 to the simplified layout.

PR #237, opened 2026-02-05
Try maxmind geoip 5.x version

mashhurs:
FYI: since moving to 5.x is a breaking change, plugin ugraded maxmind dependencies to 4.x version - https://github.com/logstash-plugins/logstash-filter-geoip/pull/238

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -37,23 +37,28 @@ plugins {
   id "base"
 }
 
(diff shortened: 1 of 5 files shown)


mashhurs:
Closing as addressed by https://github.com/logstash-plugins/logstash-filter-geoip/pull/238

PR #238, opened 2026-02-05
Upgrade maxmind geoip dependencies

### LINKED ISSUE #227: Update `geoip2` and `maxmind-db` versions
New versions of `geoip2` and `maxmind-db` were releases.
The  maxmind-db 3.1.1 comes with an NPE fix related to #226.

Update both dependencies to the latest.

Related:
- https://github.com/logstash-plugins/logstash-filter-geoip/issues/226#issuecomment-2363878092

mashhurs:
### Description
Upgrades maxmind `geoip2` to 4.4.0 and `maxmind-db` version to 3.1.1.

_**Why specifically to these versions because `geoip2` 5.x available?**_
- 5.x has breaking change(s). So far I know `dma_code` (metro code) is removed where LS is using. See draft PR for PoC - https://github.com/logstash-plugins/logstash-filter-geoip/pull/237
- aligns with Elasticsearch ingest-geoip module versions. It is not a requirement but nice to align
  - ES `8.x`, `9.x` and `main` branches use `geoip2` to 4.2.1 and maxmind-db` 3.1.1 versions.

### How to test?

!IMPORTANT - Tests are vibe coded but reviewed details.
Repo: https://github.com/mashhurs/geoip-compatibility-tests

Pull this PR and pull the [test repo](https://github.com/mashhurs/geoip-compatibility-tests).
Set the `GEOIP_PLUGIN_DIR` with the path you have closed this PR in https://github.com/mashhurs/geoip-compatibility-tests/blob/f5e126713807e89873bc8f12dfa32f41cfcde7b4/run-tests.sh#L8

I have also pushed the recent result with 8.19, 9.3 and cross (ES 8.18 vs LS 9.3 or vice versa) - https://github.com/mashhurs/geoip-compatibility-tests/tree/main/results

```
// usage
./run-tests.sh help

Usage: ./run-tests.sh [quick|8.19|9.3|cross]

Modes:
  quick - Build and run unit tests only (no Docker)
  8.19  - Full test with ES/LS 8.19 (includes elastic_integration + data streams)
  9.3   - Full test with ES/LS 9.3 (includes elastic_integration + data streams)
  cross - Cross-version compatibility tests (all Docker services)

Tests include:
  - GeoIP 4.x plugin build and unit tests
  - ES GeoIP ingest processor
  - Logstash geoip filter (upgraded plugin)
  - elastic_integration filter with data streams
```


build.gradle
```diff
@@ -40,20 +40,22 @@ plugins {
...
-String junitVersion = '5.11.2' // at least 5.11 is needed to use @FieldSource with @ParameterizedTest
-String maxmindGeoip2Version = '2.17.0'
-String maxmindDbVersion = '2.1.0'
+String maxmindGeoip2Version = '4.4.0'
+String maxmindDbVersion = '3.1.1'
+String jrubyCompleteVersion = '9.4.9.0'
+
 String log4jVersion = '2.17.1'
-String jrubyCompleteVersion = '9.1.13.0'
+String junitVersion = '5.11.2' // at least 5.11 is needed to use @FieldSource with @ParameterizedTest
```
mashhurs:
```suggestion
String junitVersion = '5.11.2'
```

src/main/java/org/logstash/filters/geoip/GeoIPFilter.java
```diff
@@ -450,7 +451,7 @@ private Map<Field, Object> retrieveAsnGeoData(InetAddress ipAddress) throws GeoI
           geoData.put(Field.IP, ipAddress.getHostAddress());
           break;
         case AUTONOMOUS_SYSTEM_NUMBER:
-          Integer asn = response.getAutonomousSystemNumber();
+          final Long asn = response.getAutonomousSystemNumber();
```
mashhurs, line 454: I was thinking if this `Integer` -> `Long` breaks the backward compatibility if result is saved into ES. Quick test with 8.19.last and 9.x versions seems by default numeric will be mapped as `Long` and this doesn't break unless if there is a strict `Integer` mapping.

```
POST /asn-type-test/_doc/1
{
  "ip": "8.8.8.8",
  "asn": 15169,
  "org": "Google LLC"
}

GET /asn-type-test/_mapping
// Results
{
  "asn-type-test": {
    "mappings": {
      "properties": {
        "asn": {
          "type": "long"
        },
        "ip": {
          "type": "text",
          "fields": {
            "keyword": {
              "type": "keyword",
              "ignore_above": 256
            }
          }
        },
        "org": {
          "type": "text",
          "fields": {
            "keyword": {
              "type": "keyword",
              "ignore_above": 256
            }
          }
        }
      }
    }
  }
}

POST /asn-type-test/_doc/2
{
  "ip": "1.2.3.4",
  "asn": 4200000000,
  "org": "Test Large ASN"
}

GET /asn-type-test/_search
```

src/main/java/org/logstash/filters/geoip/GeoIPFilter.java
```diff
@@ -236,7 +237,7 @@ private Map<Field,Object> retrieveCityGeoData(InetAddress ipAddress) throws GeoI
     CityResponse response;
     try {
       response = databaseReader.city(ipAddress);
-    } catch (NullPointerException e) {
+    } catch (NullPointerException | DeserializationException e) {
```
mashhurs, line 240: FYI: after 4.x, on top of NPE there might be `DeserializationException` if there is invalid custom field data that can't be deserialized.

build.gradle
```diff
@@ -40,20 +40,22 @@ plugins {
 group "org.logstash.filters"
 version "${new File("VERSION").text.trim()}"
 
-String junitVersion = '5.11.2' // at least 5.11 is needed to use @FieldSource with @ParameterizedTest
-String maxmindGeoip2Version = '2.17.0'
-String maxmindDbVersion = '2.1.0'
+String maxmindGeoip2Version = '4.4.0'
```
mashhurs, line 43: geoip2 interface started using maxmind-db 3.1.1 version from 4.2.1 and using the latest seems safe to me, reference - https://github.com/maxmind/GeoIP2-java/releases/tag/v4.2.1

VERSION
```diff
@@ -1 +1 @@
-7.3.3
\ No newline at end of file
+7.4.0
```
mashhurs: Did minor version upgrade since maxmind deps also upgraded to minors.

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## 7.4.0
+  - Upgrade maxmind `geoip2` to 4.4.0 and `maxmind-db` to 3.1.1 versions [#238](https://github.com/logstash-plugins/logstash-filter-geoip/pull/238)
```
mashhurs: Hm., maybe worth to mention about `asn` type (Integer to Long) change if it break anything?

build.gradle
kaisecheng: [geoip2 4.x](https://mvnrepository.com/artifact/com.maxmind.geoip2/geoip2/4.4.0/dependencies)  removed the dependency of apache httpclient and commons-codec. We can removed them too.

kaisecheng: From the maxmind's changelog, starting from 4.0.1, `com.maxmind.db` is a transitive dependency of `com.maxmind.geoip2`. How about we drop the dependency?

CHANGELOG.md
kaisecheng: Ruby only has Integer (BigInteger) type and your test shows the output is the same type, so my voice would be not mentioning here

VERSION
kaisecheng: This version drops support of jvm 8. I consider it as breaking change and suggest to change the version to 8.0.0.

build.gradle
```diff
@@ -90,6 +92,8 @@ dependencies {
   testImplementation group: "org.apache.logging.log4j", name: "log4j-api", version: log4jVersion
   testImplementation group: 'org.jruby', name: 'jruby-complete', version: jrubyCompleteVersion
   testImplementation group: "org.mockito", name: "mockito-core", version: mockitoVersion
+  // Jackson dependencies required by GeoIP2 4.x for JSON serialization (excluded above, needed for tests)
+  testImplementation group: 'com.fasterxml.jackson.core', name: 'jackson-databind', version: '2.17.2'
```
kaisecheng: I seem to be able to run test without this dependency. Which test need this?

kaisecheng (commented):
I have some questions about the dependency. We can take this chance to clean up the test cases for LS7 ([example](https://github.com/logstash-plugins/logstash-filter-geoip/blob/v7.3.3/spec/filters/geoip_spec.rb#L55-L57))

I have tested a few scenarios. All look good. 
- install in LS9 and query db
- use this plugin to query old db and custom field db
- install in LS7 and got rejected as expected

The plugin drops support on jvm 8. I think it is a breaking change. WDYT?

build.gradle
mashhurs: I hit with `java.lang.NoClassDefFoundError: com/fasterxml/jackson/databind/Module` error when running unit tests (`./gradlew clean vendor test`) if we don't have this declaration.

build.gradle
mashhurs: Ah good found! I didn't realize that artifacts API started pulling it.

mashhurs: Removed.

VERSION
mashhurs: Hm., for the given LS 7.x EOL and LS 8.x+ requires Java 11+, do we still consider it breaking change?

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## 7.4.0
+  - Upgrade maxmind `geoip2` to 4.4.0 and `maxmind-db` to 3.1.1 versions [#238](https://github.com/logstash-plugins/logstash-filter-geoip/pull/238)
```
mashhurs:
```suggestion
  - Upgrade maxmind `geoip2` to 4.4.0 version [#238](https://github.com/logstash-plugins/logstash-filter-geoip/pull/238)
```

VERSION
kaisecheng: LS7 is EOL but still it is an open source product that is free to run in production and widely use based on product telemetry. From the user's point of view, they are using 7.3.4 with LS7 and suddenly they are not able to run 7.4.0. I think this break customer's expectation.

We can ask team to give input

build.gradle
kaisecheng: Strange, I am not able to replicate the failure. I created a test pr to comment out the line and all tests [passed](https://app.travis-ci.com/github/logstash-plugins/logstash-filter-geoip/builds/277457692). Can you double check it?

kaisecheng (commented):
Can you clean up the unit tests related to LS 7 for [example](https://github.com/logstash-plugins/logstash-filter-geoip/blob/v7.3.3/spec/filters/geoip_spec.rb#L55-L57)?

build.gradle
mashhurs: Hm., it looks like `./gradlew clean` doesn't remove the vendor folder (where fasterxml JAR is located). Can you please manually remove vendor and re-run the `./gradlew clean vendor test`?

VERSION
mashhurs: I cannot disagree with your point.
We were discussing this with @jsvd in our 1:1 and the fact that we bundle JDK in LS core and aligning with it would give us a better support. There are also situations encourage to move fast. One of them is move to jruby  10 (sunset is coming to 9) which requires Java 25 (João was mentioning even we might need to backport to 8.x omg!).
So, overall we don't have concerns!

mashhurs:
> Can you clean up the unit tests related to LS 7 for [example](https://github.com/logstash-plugins/logstash-filter-geoip/blob/v7.3.3/spec/filters/geoip_spec.rb#L55-L57)?

Done!

build.gradle
mashhurs: With your test CI PR it looks like java tests are passing w/o issue - https://github.com/logstash-plugins/logstash-filter-geoip/pull/240/checks?check_run_id=63391316821
It might be my env issue that I even cannot run `./.ci/docker-setup.sh && ./.ci/docker-run.sh` somehow (error below). Removed fasterxml test dependency.

```
logstash-1  | > Task :compileJava FAILED
logstash-1  | /usr/share/plugins/plugin/src/main/java/org/logstash/filters/geoip/GeoIPFilter.java:37: error: cannot find symbol
logstash-1  | import org.logstash.Event;
```

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## 7.4.0
+  - Upgrade maxmind `geoip2` to 4.4.0 version [#238](https://github.com/logstash-plugins/logstash-filter-geoip/pull/238)
```
mashhurs:
```suggestion
## 8.0.0
  - Upgrade maxmind `geoip2` to 4.4.0 version [#238](https://github.com/logstash-plugins/logstash-filter-geoip/pull/238)
    - Dropped support for Logstash 7.x. Minimum supported version is now Logstash 8.0.0
    - Dropped support for Java 8. Minimum required version is now Java 11
```

VERSION
```diff
@@ -1 +1 @@
-7.3.4
\ No newline at end of file
+7.4.0
```
mashhurs:
```suggestion
8.0.0
```

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 8.0.0
+  - Upgrade maxmind `geoip2` to 4.4.0 version [#238](https://github.com/logstash-plugins/logstash-filter-geoip/pull/238)
+    - Dropped support for Logstash 7.x. Minimum supported version is now Logstash 8.0.0
+    - Dropped support for Java 8. Minimum required version is now Java 11
```
kaisecheng:
```suggestion
  - Dropped support for Logstash 7.x. Minimum supported version is now Logstash 8.0.0
  - Dropped support for Java 8. Minimum required version is now Java 11
```

kaisecheng (approved):

kaisecheng (approved):
LGTM. Thank you for upgrading geoip 🚀

